PR #7, opened 2021-02-22
Add contributing info and spec status definitions

shannonwells:
Problem
=======
Adding a Contributing page is of the tasks in
[#177008925](https://www.pivotaltracker.com/story/show/177008925)

Solution
========
I added a short contributing section to the README.  I debated on whether to include it in the spec website "pages" folder, however, this is information aimed at spec writers, so I decided it probably belongs at the top level as with most other repos with this sort of page.  It seemed silly to create a special CONTRIBUTING page with just this short amount of info.  Happy to discuss it though.

The second thing I did was to add a spec status definition to the Welcome page. I'm not at all sure about these definitions so I would welcome feedback on this.

Whichever gets merged first, this or Archive messages spec, I'll add this diagram (or its accepted revision) at the bottom:
Status-process


enddynayn:
👍

README.md
```diff
@@ -14,3 +14,27 @@ npm run dev
...
+* The initial spec status is _Draft_.  Please see the diagram below for the specification lifecycle.
+* Choose a target version number of the platform for the spec.
+* Use the official outline for specifications
+
+## Spec status definitions
+
+| Name | Description |
+| --- | --- |
+| Draft | Spec is posted as draft PR. It's open for comment and major changes. |
+| Proposed | Ready for formal review. It should be done except for minor changes. |
```
wilwade, line 31: I think the Draft status is throwing me off with "PR" in it. To me it would imply that nothing is merged until it is at least to the Proposed status. I wonder which option is desired (or intended):
1.  PR is implied as before Draft and "PR" should be removed from Draft
2. We need something between Draft and Proposed
3. Or the intent is that nothing should be merged and have longer running branches around a spec.

wilwade (commented):
Just one comment. The framing is really nice.

README.md
shannonwells: The process I envisioned goes like this:

1. Spec writer(s) post a GitHub "draft PR" with their draft of a spec.  This spec is open for lengthy discussion, major changes if needed.
2. Once the discussions have been resolved by the spec writer(s), they take the PR out of "draft" and the status to "proposed".  The only changes that should be occurring by this point are minor ones, like reorganizing a section, clarifications, language errors.
3. Just before the spec is merged to main, it should be changed to "tentative."
4. Once the spec has been solidified, i.e. we've built pretty much everything and we know things aren't going to change -- probably once it reaches v 1.0, the spec status should be changed to final.

Once the spec is in "Tentative", any major changes require a new spec process and a bumped version once accepted as "tentative."

I realize that this spec status is a little bit redundant with GitHub's functionality, but if we were to post these docs somewhere other than GitHub people will know what things mean.

wilwade: Hmm... That process does sound good. I agree with the need to keep it separate from GitHub's functionality.

Nothing in a PR however would be viewable on the website version of the spec. I imagined that we would want to be able to show new draft specs there, but perhaps not?

I wonder how this works for modifications for existing spec? If I want to modify Spec X do I put the status back into draft?

I think your description of the flow above (or whatever we end up with) should be added to the contributing document. The spec status definitions are great from the perspective of understanding what they mean for a reader, but the description above is much better for how to process them as a spec writer.

@rosalinekarr I wonder if we can add a lint that will check for "Tentative" or "Final" in the spec so we cannot merge without that (or whatever the rule ends up being)

shannonwells: For modifications of an existing spec:
1.  Leave the spec as-is in main .
2. Follow the process above through step 3, except bump the version as appropriate. 
3. Once it's merged, the changes should reflect "Tentative" status, + the new version number.
4. Proceed with step 4 as before, as appropriate.

shannonwells: > Nothing in a PR however would be viewable on the website version of the spec. I imagined that we would want to be able to show new draft specs there, but perhaps not?

Eh, I guess it's just in case?  I mean there will be a link to the (public) GitHub, so whomever is looking at the spec and maybe they want to suggest something, they can come to the repo & look at PRs.

wilwade: Sounds good to me. We'll revise as needed.

README.md
```diff
@@ -14,3 +14,59 @@ npm run dev
 ```
 
 Once the local server is running, a live-reloading preview of the spec can be viewed at http://localhost:3000.
+
+## Tools and frameworks used in this repo
+* [Dokz](https://dokz.site/docs)
+* [MDX](https://mdxjs.com/)
+* [Chakra-UI](https://chakra-ui.com/docs/getting-started)
```
wilwade, line 21: Some of these might change with https://github.com/LibertyDSNP/spec/pull/14

README.md
```diff
@@ -14,3 +14,59 @@ npm run dev
...
+1. The initial spec status is _Draft_.  Please see [Process](#Process) for details.
+1. Choose a target version number of the platform for the spec.
+1. Use the official outline for specifications1. Respect and observe good file organization practices. For example, if your spec relates to DSNP please put the spec in the _DSNP_ folder and not anywhere else.
+1. Spec files are written in markdown (.md) format, and are located in the _pages_ directory.
+1. Use dashes (`-`) between words in file names so that they appear in the Table of Contents correctly.  Use MDX format (`.mdx` extension) if you want to import React components in the page.
+1. When documenting an API, et, use the following field template:
+
+   | name | description | type | required?|
+   | ---|---|---|---|
+   | thisThing | a description of thisThing | bytes | YES |
```
wilwade, line 35: Be nice to get consistent on our spacing, but that's more of a linting question which is already in progress.

wilwade (approved):

PR #8, opened 2021-02-22
DSNP Announce Spec

wilwade:
Problem
=======

Solution
========

A separate interface for creating an DSNPMessage event that announces a new message.

Change summary:
---------------
* Added to the sidebar
* Referenced DSNPMessage

### DIFF
--- a/pages/DSNP/DSNP-Announce.md
+++ b/pages/DSNP/DSNP-Announce.md
@@ -31,7 +31,7 @@ interface DSNPAnnounce {
 
 | field | description | type |
 |-------|-------------|------|
-| dnspFrom | Social Identity From Address | address / uint160 / bytes20 |
+| dnspFrom | Social Identity From Address | address |
 | dsnpType | Enum from [DSNP Message Types](/DSNP/DSNP-Message-Types) | unit16 |
 | data | Matching serialized data from [DSNP Messages](/DSNP/DSNP-Messages) | bytes |
 
@@ -39,7 +39,7 @@ interface DSNPAnnounce {
 
 The event topic for DSNPMessage follows the standard Solidity event name to hash standard.
 ```
-0x98aa3317938b163d794c4cf04dce050dbf5b96a939a7387b4b7c38f3cafcbf06 = keccak-256("DSNPMessage(from,uint16,bytes32)")
+0xd4248342bca04dd21d5e78bc87bf14992ef919be52514fc229e12a9977949cb9 = keccak-256("DSNPMessage(address,uint16,bytes32)")
 ```
 
 ### Announce Requirements
--- a/pages/DSNP/DSNP-Announce.md
+++ b/pages/DSNP/DSNP-Announce.md
@@ -1,0 +1,62 @@
+# DSNP Announce
+## Specification Status
+
+| Version | Status |
+---------- | ---------
+| 0.1     | Draft |
+
+## Purpose
+1. Describe the method to announce [DSNP Messages](/DSNP/DSNP-Messages) to the blockchain.
+1. Provide Solidity code examples (pending final contract)
+1. Facilitate use of SDK and locating of on-chain data
+
+## Assumptions
+* Chain messages are on Ethereum
+* Message data is posted via [Ethereum log events](https://medium.com/mycrypto/understanding-event-logs-on-the-ethereum-blockchain-f4ae7ba50378)
+* Signature algorithm is [secp256k1](https://en.bitcoin.it/wiki/Secp256k1). This allows the use `ecreover`
+  to get public keys. A public key also need not be included in a log event for ease of validation.
+* content hashes are created via the same [keccak-256 hashing algorithm](https://en.wikipedia.org/wiki/SHA-3) used by Ethereum.
+
+## DSNP Announce
+
+### DSNPMessage Event
+
+The DSNPMmessage event is a standard event topic to be used for all DSNP Messages.
+
+```solidity
+interface DSNPAnnounce {
+    event DSNPMessage(address from, uint16 dsnpType, bytes32 data);
+}
+```
+
+| field | description | type |
+|-------|-------------|------|
+| from | Social Identity From Address | address / uint160 / bytes20 |
+| dsnpType | Enum from [DSNP Message Types](/DSNP/DSNP-Message-Types) | unit16 |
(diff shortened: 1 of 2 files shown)


pages/DSNP/DSNP-Announce.md
```diff
@@ -0,0 +1,62 @@
...
+
+```solidity
+interface DSNPAnnounce {
+    event DSNPMessage(address from, uint16 dsnpType, bytes32 data);
+}
+```
+
+| field | description | type |
+|-------|-------------|------|
+| from | Social Identity From Address | address / uint160 / bytes20 |
```
wilwade, line 34: 1. Should we have this here?
2. Should we name it something different?
3. Should it use the Ethereum address type or do we want to make it generic bytes?

pages/DSNP/DSNP-Announce.md
```diff
@@ -0,0 +1,62 @@
...
+| dsnpType | Enum from [DSNP Message Types](/DSNP/DSNP-Message-Types) | unit8 |
```
wilwade, line 35: `type` is a reserved word. I don't really like `dsnpType`. Other ideas?

pages/DSNP/DSNP-Announce.md
```diff
@@ -0,0 +1,62 @@
...
+| data | Matching serialized data from [DSNP Messages](/DSNP/DSNP-Messages) | bytes |
```
wilwade, line 36: Naming is hard. Over in the messaging spec this is referred to as `dsnpData`, but that feels redundant in the context of a DSNPMessage event.

shannonwells (approved):
YUSS

pages/DSNP/DSNP-Announce.md
```diff
@@ -0,0 +1,62 @@
...
+```solidity
+interface DSNPAnnounce {
+    event DSNPMessage(address from, uint16 dsnpType, bytes32 data);
+}
+```
+
+| field | description | type |
+|-------|-------------|------|
+| from | Social Identity From Address | address / uint160 / bytes20 |
+| dsnpType | Enum from [DSNP Message Types](/DSNP/DSNP-Message-Types) | unit16 |
```
wilwade, line 35: 1. `type` is a reserved word. I don't really like `dsnpType`. Other ideas?
2. Is uint16 enough?

pages/DSNP/DSNP-Announce.md
shannonwells: @rosalinekarr and I chatted about this a little and agreed that naming is hard. I chose dsnp___ because even though it's stuttery, when it's viewed within a documentation context it will never be confusing as to what it is.

pages/DSNP/DSNP-Announce.md
shannonwells: dsnpAction was another name considered.  or dsnpMessageType, or messageType.

pages/DSNP/DSNP-Announce.md
shannonwells: but similarly to above, I also considered "data" and "messageData"  I just remembered, picked dsnp___ also because it's shorter than message.

pages/DSNP/DSNP-Announce.md
shannonwells: I happen to like the convention either of `from` or `fromAddr` as a name.    

If it's generic bytes, then how does it get mapped to a social identity contract?

pages/DSNP/DSNP-Announce.md
shannonwells: I think `uint16` is more than enough.

pages/DSNP/DSNP-Announce.md
wilwade: I wonder how `dsnp_____` holds up against other `dsnp` events we will have, but I think it works for now. We can change it later without it effecting the log event topic (which is signature based, but does not include the parameter names)

pages/DSNP/DSNP-Announce.md
wilwade: I tried out the name `dsnpFrom` and standardized the prefixes. Thoughts?

In Solidity address, uint160, bytes20 are all synonyms. The connection would likely come as part 2. The identity interface likely will be required to implement https://eips.ethereum.org/EIPS/eip-1271 that would allow for verification at the level of the announcing contract and the end user.

Likely that will also require that the signature (bytes32) be included as a core log level data point.

pages/DSNP/DSNP-Announce.md
wilwade: After thinking it over I think I like the dsnp prefixes (and have perhaps tried to over use them now.)

pages/DSNP/DSNP-Announce.md
wilwade: After thinking it over I think I like the dsnp prefixes (and have perhaps tried to over use them now.)

pages/DSNP/DSNP-Announce.md
```diff
@@ -0,0 +1,62 @@
...
+
+```solidity
+interface DSNPAnnounce {
+    event DSNPMessage(address dnspFrom, uint16 dsnpType, bytes[] dnspData);
+}
+```
+
+| field | description | type |
+|-------|-------------|------|
+| dnspFrom | Social Identity From Address | address / uint160 / bytes20 |
```
wilwade, line 34: @shannonwells What do you think of this naming?

shannonwells: fine with me

pages/DSNP/DSNP-Announce.md
```diff
@@ -0,0 +1,62 @@
...
+### Event Topic
+
+The event topic for DSNPMessage follows the standard Solidity event name to hash standard.
+```
+0x98aa3317938b163d794c4cf04dce050dbf5b96a939a7387b4b7c38f3cafcbf06 = keccak-256("DSNPMessage(from,uint16,bytes32)")
+```
+
+### Announce Requirements
+
+| Interface | Required |
```
shannonwells, line 47: Also I like this section.  Maybe we should have a spec template too.

shannonwells (commented):
With respect to labeling address types, let's just say "address" in the type and make a note at the top about address/uint160/bytes20

Also I think it would be good to have a short description of what the Announce contract is meant for.

I think I still prefer the "route" idea but there are good non-conflicting things in here that would go in #11 .

PR #11, opened 2021-02-24
DSNP Announce Spec with Routing

wilwade:
Note
=======
This is a alternative version of https://github.com/LibertyDSNP/spec/pull/8. The versions cannot both be merged.

Problem
=======

Solution
========

A separate interface for creating an DSNPMessage event that announces a new message.

Change summary:
---------------
* Added to the sidebar
* Referenced DSNPMessage
* Added the concept of dsnpRoute
* Split Broadcast and Replies


pages/DSNP/DSNP-Announce.md
```diff
@@ -0,0 +1,102 @@
...
+
+The id of a message the new message is related to.
+Used for replies, reactions, and other message related metadata.
+If you want to gather additional messages about a message you are showing, look for this route.
+
+#### Route Field Example: Personal Id
+
+Messages that are for a user and their applications only.
+Usually contains encrypted data.
+Generated by taking the keccak-256 hash of the user's id.
```
wilwade, line 75: I'm not sure what I think of this still. Both as a concept and this implementation. Also I'm not sure about the list of things I applied to it.

I like the idea of a `hash(sender's id)` meaning that personal stuff is less obviously connected. It is NOT private however, so perhaps that is just not worth the work?

Another option is just to prepend a 1 to the sender's id (only works when the sender's id is < 32 bytes which it is currently).
Another option is to just add 1 to the sender's id (only works if the ids are in a large space, which they are currently).

pages/DSNP/DSNP-Announce.md
```diff
@@ -0,0 +1,102 @@
...
+
+The id of a message the new message is related to.
+Used for replies, reactions, and other message related metadata.
+If you want to gather additional messages about a message you are showing, look for this route.
+
+#### Route Field Example: Personal Route Id
+
+Messages that are for a user and their applications only.
+Usually contains encrypted data.
+Generated by taking the keccak-256 hash of the user's id.
```
wilwade, line 75: I'm not sure what I think of this still. Both as a concept and this implementation. Also I'm not sure about the list of things I applied to it.

I like the idea of a `hash(sender's id)` meaning that personal stuff is less obviously connected. It is NOT private however, so perhaps that is just not worth the work?

Another option is just to prepend a 1 to the sender's id (only works when the sender's id is < 32 bytes which it is currently).
Another option is to just add 1 to the sender's id (only works if the ids are in a large space, which they are currently).

pages/DSNP/DSNP-Message-Types.md
```diff
@@ -17,20 +17,21 @@
...
-| EncryptionKeyList | Encryption keyList rotation | 6 |
-| Reaction | a public visual reply to a Broadcast | 7 |
-| PrivateGraphChange | an encrypted follow or unfollow | 8
-| Drop | a dead drop message | 9
-| EncryptedInbox | an encrypted direct message | 10
-| PrivateBroadcast | an encrypted broadcast | 11
+| name     | route | description | value |
+|-------   |-------------|-------------| ----|
+| Private | Personal Route ID | message and message type are private; message data is encrypted | 0 |
+| GraphChange | Personal Route ID | follow or unfollow an account | 1 |
```
wilwade, line 23: I'm not sure what the route should be for public graph changes (assuming of course that graph stays here). Thoughts?

shannonwells: Why wouldn't it just be the social identity for the graph change?

Also, I think if we're going to keep "Private" then we don't need a PrivateGraphChange type.  But I can change that in my PR.

pages/DSNP/DSNP-Messages.md
```diff
@@ -62,19 +62,27 @@ a public post (was Announcement)
...
 
-**NOTE** If origin broadcasts and replies stay the same type, inReplyTo is allowed to be blank.  If replies are separated into their own type, the inReplyTo field here will be dropped.
+#### Reply
+a public reply post
+
+| field     | description | type |
+|-------    |-------------| ----|
+| dnspRoute (inReplyTo) | messageID replied to | bytes32
+| messageID | keccak-256 hash of content stored at uri |  bytes32
+| fromAddress | sender's identity | bytes20
```
shannonwells, line 76: `fromAddress` isn't needed in the actual DSNP data segment.

pages/DSNP/DSNP-Messages.md
```diff
@@ -62,19 +62,27 @@ a public post (was Announcement)
 
 | field     | description | type |
 |-------    |-------------| ----|
-| inReplyTo | messageID replied to | bytes32
+| dnspRoute (fromAddress) | id of the sender | bytes20
```
shannonwells, line 65: I thought the idea here was to remove this field from the DSNP data so it wouldn't be duplicated.

shannonwells (changes requested):
Overall I like the route idea, but this PR needs some clarification - there are some things that contradict what you said yesterday.

pages/DSNP/DSNP-Messages.md
wilwade: It is. I don't like how I am indicating it. The idea is to note what data the dsnpRoute contains, but you are right that it doesn't look that way. I'll work on a better way to show that information.

pages/DSNP/DSNP-Messages.md
wilwade: I think it is. Where else would that data come from?

pages/DSNP/DSNP-Messages.md
wilwade: Here are three different ways to show the data. Not sure which I like:

image

pages/DSNP/DSNP-Messages.md
shannonwells: the fromAddress of the message?

wilwade: That assumes that the message is coming from the identity contract. If this is a separate contract, then it doesn't work. I think I added it to all the ones that need it. Most of then actually don't. (Drop for instance)

pages/DSNP/DSNP-Messages.md
wilwade: @shannonwells Updated to a new style

pages/DSNP/DSNP-Message-Types.md
wilwade: Re: Private vs PrivateGraphChange. I think there is a lot of future discussion around this. Perhaps even a PR that merges them along with KeyList even

The idea behind the Personal Route ID is to note that it is not for routing to the public. These are not the messages you want if you want messages from a given person.

shannonwells (approved):

PR #14, opened 2021-02-25
Setup Deployment

rosalinekarr:
Problem
=======
We want to automatically deploy to GitHub Pages
[#176962224](https://www.pivotaltracker.com/story/show/176962224)

Solution
========
Added GitHub Actions to auto-deploy to `gh-pages` branch

with @enddynayn


.github/workflows/main.yml
```diff
@@ -0,0 +1,22 @@
...
+    branches:
+      - main
+jobs:
+  build-and-deploy:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout 🛎️
+        uses: actions/checkout@v2.3.1
+
+      - name: Install and Build
```
wilwade, line 13: Do we want / need to cache? https://github.com/actions/cache/blob/main/examples.md#node---npm

.github/workflows/main.yml
```diff
@@ -0,0 +1,22 @@
+name: Build and Deploy
+on:
+  push:
+    branches:
+      - main
+jobs:
+  build-and-deploy:
+    runs-on: ubuntu-latest
```
wilwade, line 8: Can we specify which node version to use?

package.json
```diff
@@ -2,22 +2,27 @@
...
-        "@emotion/react": "11.1.5",
-        "@emotion/styled": "11.1.5",
-        "framer-motion": "3.3.0",
-        "dokz": "2.0.5",
-        "next": "10.0.7",
-        "react": "17.0.1",
-        "react-dom": "17.0.1"
+        "@emotion/react": "^11.1.1",
+        "@emotion/styled": "^11.0.0",
+        "docz": "latest",
```
wilwade, line 20: We might want to specify the version here to limit package-lock churn.

package.json
```diff
@@ -2,22 +2,27 @@
...
+        "react": "^16.8.6",
```
wilwade, line 21: We don't get 17? :(

doczrc.js
```diff
@@ -0,0 +1,47 @@
+export default {
+  base: "/spec/",
+  typescript: true,
+  ignore: ["CODEOFCONDUCT.md", "LICENSE.md", "README.md"],
+  menu: [
```
wilwade, line 5: I recall that the other setup auto added to this. Do new files get auto added if running in dev mode? If not do we need to add this to the readme as a note to add stuff?

wilwade (changes requested):
Few notes, but looking good.

.github/workflows/main.yml
rosalinekarr: We probably don't need to, but I can go ahead and add it. Currently, our workflow runs in just under half a second.

doczrc.js
rosalinekarr: New files will automatically be added, but if you want to control where they appear in the sidebar, we'll have to add them here. Dokz actually did the same thing, but it buried this config in `pages/.app.jsx`.

package.json
```diff
@@ -2,22 +2,27 @@
...
-        "react-dom": "17.0.1"
+        "@emotion/react": "^11.1.1",
+        "@emotion/styled": "^11.0.0",
+        "docz": "^2.3.1",
+        "react": "^17.0.1",
+        "react-dom": "^16.8.6"
     },
     "devDependencies": {
-        "compose-function": "3.0.3"
+        "@types/react": "^16.8.23",
```
wilwade, line 25: React / Type version mismatch.

.github/workflows/main.yml
```diff
@@ -0,0 +1,33 @@
+name: Build and Deploy
+on:
+  push:
+    branches:
+      - main
+jobs:
+  build-and-deploy:
+    runs-on: ubuntu-16.04
```
wilwade, line 8: Isn't 16.04 EOL'd?

package.json
```diff
@@ -2,22 +2,27 @@
...
-        "framer-motion": "3.3.0",
-        "dokz": "2.0.5",
-        "next": "10.0.7",
-        "react": "17.0.1",
-        "react-dom": "17.0.1"
+        "@emotion/react": "^11.1.1",
+        "@emotion/styled": "^11.0.0",
+        "docz": "^2.3.1",
+        "react": "^17.0.1",
+        "react-dom": "^16.8.6"
```
wilwade, line 22: Should this match the react version?

wilwade (dismissed):
Three more questions.

package.json
rosalinekarr: fixed

.github/workflows/main.yml
rosalinekarr: changed to 20.04

package.json
rosalinekarr: fixed

enddynayn:
👍

PR #15, opened 2021-02-26
DSNP Annoucing Spec

wilwade:
Note
=======
This is a alternative version of https://github.com/LibertyDSNP/spec/pull/8 and https://github.com/LibertyDSNP/spec/pull/11.
The versions cannot all be merged.

Problem
=======

Solution
========

A separate interface for creating an DSNPBatch event that announces a file that contains the list of new messages.

Change summary:
---------------
* Added to the sidebar
* Referenced DSNPMessage
* DSNPAnnounce Interface
* DSNPBatch event


pages/DSNP/Messages/Overview.md
```diff
@@ -65,94 +64,112 @@ For details on how messages are serialized, see [DSNP Message Serialization](/DS
...
 
 #### Drop
 a dead drop message
 
-| field | description | type
-|-------|-------------| ---|
-| ddid | dead drop id |  bytes
-| uri  | content uri  |  bytes
+| dsnpData field | description | type |
+| ------------- |------------- | ---- |
```
shannonwells, line 89: I learned recently you just need three dashes between pipes to cause it to be rendered as a table.

shannonwells (commented):
Will this be irrespective of whether routes are used?

pages/DSNP/Messages/Overview.md
wilwade: @shannonwells I wonder if we want a lint formatter to do the table formatting for us... Make them look right every time. Or I am also ok with a standard of just three `---` and spaces before and after. So like:
```markdown
| title | spaces around |
| --- | --- |
| content | also has spaces around |
```

wilwade:
@shannonwells Re: Routes. That is correct. This is independent of routes, and routes should be re-evaluated in the context of the batch file.

shannonwells (approved):
ell gee tee em

PR #17, opened 2021-03-01
glossary from admin wiki added

claireolmstead:
Problem
=======
As a developer reading the spec I want a glossary.  So I can look up unfamiliar terms when I encounter them[link to Pivotal Tracker #177121493](https://www.pivotaltracker.com/story/show/177121493)

Solution
========
* Copy over the existing content from the admin wiki glossary into the glossary reference file on the Spec.
* Ask the blockchain team if there are any terms to be added or omitted from the list. 

Change summary:
---------------
* added content to the glossary page.

Steps to Verify:
----------------
1. Check that the list of terms exists and has the desired content.

Screenshots (optional):
-----------------------
Show-n-tell images/animations here


pages/Reference/glossary.md
```diff
@@ -4,4 +4,24 @@ route: /Reference/Glossary
 menu: Reference
 ---
 
-TBD
+# Glossary
+
+* **Account** - See "Social Identity"
+* **Announcement** - A item of data, typically an image or note, either directly posted or hashed then posted to the blockchain
```
rosalinekarr, line 10: "A item" => "An item"

rosalinekarr: Also, "either directly posted or hashed then posted to the blockchain" might not be accurate anymore. Maybe we should just replace that bit with "posted to the blockchain" without any details?

pages/Reference/glossary.md
```diff
@@ -4,4 +4,24 @@ route: /Reference/Glossary
...
+* **Announcer** - The contract that posts new content from a publisher to the blockchain through batches or batch announcements by an archivist
+* **Publisher** - A user or bot who posts new content to an announcer
+* **Bad actor** - A user intentionally using the application for bad-faith or illegal purposes
+* **Bot** - An automated account, sometimes malicious but often providing some service ("Hey Alexa, what's the weather?")
+* **Consumer** - Someone who reads content from social media
+* **Contract address** - The unique number associated with a smart contract posted on the blockchain
+* **Happy path** - The expected user flow assuming no errors or bad actors
```
rosalinekarr, line 17: Let's drop this one too. It's more of any industry term than a project-specific thing.

rosalinekarr (changes requested):
Just a couple of notes, otherwise LGTM

rosalinekarr (approved):

PR #20, opened 2021-03-04
Chore/uris to strings

shannonwells:
Problem
=======
So that the spec will match what we are planning for contracts in Ethereum/Solidty, update URIs in the spec to be type `string`
[Fixes #177208548](https://www.pivotaltracker.com/story/show/177208548)

Solution
========
Update field type bytes to string anywhere there is a  URI.  I changed it in Archive Messages too just for ease of conversion.

with @pairperson1

Change summary:
---------------
* Changed bytes --> string in Archive Messages and DSNP Messages
* Bumped version
* Updated Status according to process.

Steps to Verify:
----------------
Please make sure I didn't miss any places where a URI field should be a string.

### DIFF
--- a/.spelling
+++ b/.spelling
@@ -56,3 +56,5 @@ toAddress
 utf-8
 ActivityPub
 Braxton
+Alexa
+prosumer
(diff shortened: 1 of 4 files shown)


wilwade (approved):
🥇

PR #25, opened 2021-03-22
Fix links and make links match filenames

wilwade:
Problem
=======
problem statement, including
[link to Pivotal Tracker #177439791](https://www.pivotaltracker.com/story/show/177439791)

Solution
========
Updated links to make sure they all worked.

Change summary:
---------------
* Updated broken links
* Updated link for /implementation_status to /Implementation_Status as github pages is case sensitive and now case matches filenames

Steps to Verify:
----------------
1. Run a local instance
1. Check internal links


aneyzberg (approved):
👍

PR #26, opened 2021-03-22
Very short! Add some clarifying text, other serialization options

shannonwells:
Problem
=======
I realized that in our Message Serialization page, I did not specify what the serialization was _for_, specifically. This is specifically for posting messages on chain.

Solution
========
I added some explanation. I also added other serialization possibilities that we're considering.
Screenshots (optional):

### DIFF
--- a/pages/Messages/Serialization.md
+++ b/pages/Messages/Serialization.md
@@ -13,7 +13,7 @@ menu: Messages
 | 0.1     | Tentative |
 
 ## Purpose
-1. Describe the form and method of DSNP message serialization.
+1. Describe the form and method of DSNP message serialization for posting messages in Ethereum logs.
 1. Facilitate use of SDK
 
 ## Assumptions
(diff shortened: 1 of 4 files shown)


pages/Messages/Serialization.md
```diff
@@ -13,7 +13,7 @@ menu: Messages
 | 0.1     | Tentative |
 
 ## Purpose
-1. Describe the form and method of DSNP message serialization.
+1. Describe the form and method of DSNP message serialization for posting messages in Ethereum logs.
```
wilwade, line 16: Is the Eth logs or the Batch Entries? Or perhaps both?

wilwade (changes requested):
One question

pages/Messages/Serialization.md
shannonwells: It means, as opposed to serializing messages in response to queries to Batchers or Archivists.

wilwade (approved):

PR #28, opened 2021-03-26
Identity Interfaces

wilwade:
Problem
=======
problem statement, including
[link to Pivotal Tracker #177379707](https://www.pivotaltracker.com/story/show/177379707)

Solution
========

Document the spec for the IDelegation and the delegation model as well as the Factory interfaces for building identities

Change summary:
---------------
* Create Identity Overview
* Build IDelegation interface
* List additional EIP Requirements
* Identity Factory Spec

PR #30, opened 2021-04-09
Features Graph Identifier & Registry

wilwade:
Based on https://github.com/LibertyDSNP/spec/pull/28

Problem
=======
A way to have a separate graph identifier from just the contract address
[link to Pivotal Tracker #177445350](https://www.pivotaltracker.com/story/show/177445350)

Solution
========
Build up an interface for a registry, similar to ENS.

Change summary:
---------------
* Filled out the registry spec

### DIFF
--- a/.spelling
+++ b/.spelling
@@ -83,3 +83,4 @@ uint64
 ecrecover
 Punycode
 TODO
+unicode
--- a/.spelling
+++ b/.spelling
@@ -82,5 +82,6 @@ ENS
 uint64
 ecrecover
 Punycode
+punycode
 TODO
 unicode
--- a/.spelling
+++ b/.spelling
@@ -81,3 +81,5 @@ NFT
 ENS
 uint64
 ecrecover
+Punycode
+TODO
--- a/pages/Identity/Registry.md
+++ b/pages/Identity/Registry.md
@@ -28,14 +28,12 @@ and allows switching identity contracts while maintaining all graph connections,
 Contract addresses are not easy to remember and while some networks have had success with just numbers,
 it is most common to think of text based handles for finding other people on a network.
 
-## Graph Connections
+## Graph Connection Ids
 
 Ethereum contract addresses are currently 160 bit values which is much larger than needed for unique identification.
 Graph connections can be reduced to just 64 bit identifiers.
 
 ## EIP 721
-
-This needs to be evaluated and discussed for inclusion/exclusion.
 
 While the NFT standard is not necessary to support, it provides standard ownership transfer and management tools.
 
@@ -50,10 +48,20 @@ While the NFT standard is not necessary to support, it provides standard ownersh
 ## Discussion Points
 
 * Should the interface support reverse resolutions? (id -> handle, address -> id/handle)
+  * YES
 * Should we still match the ENS Resolver spec [EIP 137](https://eips.ethereum.org/EIPS/eip-137#resolver-specification)?
+  * NO, stay away so that the future can implement it if wanted
 * Should we name this registry or resolver or split?
+  * NO 
 * Do we need the uint64 id or should we just use the handle?
-* Are we going to hash the handle (ENS style) and should that happen on the client side?
+  * YES
+* Are we going to hash the handle (ENS style) client side?
+  * NO, we will store the strings
+* Can more than one handle point to the same social identity?
+  * NO
+* Are we case sensitive / normalizing?
+  * Strings are UTF-8
+  * Add a new story to investigate and a note that it is an open question
 
 ## Contract Interface
 
@@ -79,33 +87,6 @@ interface IRegistry {
     event DSNPHandleChange(address addr, uint64 id);
 
     /**
-     * @dev Resolve a handle to a contract address
-     * @param handle The handle to resolve
-     * 
-     * @throws if not found
-     * @returns Address of the contract
-     */
-    function resolveHandleToAddress(string handle) view external returns (address);
-
-    /**
-     * @dev Resolve an id to a contract address
-     * @param id The id to resolve
-     * 
-     * @throws if not found
-     * @returns Address of the contract
-     */
-    function resolveIdToAddress(uint64 id) view external returns (address);
-
-    /**
-     * @dev Resolve a handle to the id
-     * @param handle The handle to resolve
-     * 
-     * @throws if not found
-     * @returns The uint64 id for the handle
-     */
-    function resolveHandleToId(string handle) view external returns (uint64);
-
-    /**
      * @dev Register a new handle
      * @param addr Address for the handle to point at
      * @param handle The handle for the address
@@ -126,7 +107,7 @@ interface IRegistry {
      * MUST be signed by someone who is authorized on the contract via `EIP 1271(addr).isValidSignature`
      * MUST emit DSNPHandle
      */
-    function registerBySignature(bytes32 r, bytes32 s, uint32 v, address addr, string handle) external;
+    function registerByEIP712Sig(bytes32 r, bytes32 s, uint32 v, address addr, string handle) external;
 
     /**
      * @dev Alter a handle resolution
@@ -134,6 +115,7 @@ interface IRegistry {
      * @param handle The handle to alter
      * 
      * MUST be called by someone who is authorized on the contract via `EIP 1271(oldAddr).isValidSignature`
+     * TODO: FIX THE ISSUE OF newAddr not being a part of the creation
      * MUST emit DSNPHandleOwnershipChange
      */
     function setAddr(address newAddr, string handle) external;
@@ -150,5 +132,33 @@ interface IRegistry {
      * MUST emit DSNPHandle
      */
     function setAddrSignature(bytes32 r, bytes32 s, uint32 v, address newAddr, string handle) external;
+
+    /**
+     * @dev Resolve a handle to a contract address
+     * @param handle The handle to resolve
+     * 
+     * @throws if not found
+     * @returns Address of the contract
+     */
+    function resolveHandleToAddress(string handle) view external returns (address);
+
+
+    /**
+     * @dev Resolve a handle to the id
+     * @param handle The handle to resolve
+     * 
+     * @throws if not found
+     * @returns The uint64 id for the handle
+     */
+    function resolveHandleToId(string handle) view external returns (uint64);
+
+    /**
+     * @dev Resolve an id to a contract address
+     * @param id The id to resolve
+     * 
+     * @throws if not found
+     * @returns Address of the contract
+     */
+    function resolveIdToAddress(uint64 id) view external returns (address);
 }
 ```
--- a/pages/Identity/Registry.md
+++ b/pages/Identity/Registry.md
@@ -21,21 +21,43 @@ and allows switching identity contracts while maintaining all graph connections,
 
 ## Assumptions
 
-* Handles will need to be moved from Betanet to Mainnet 
+* Ids will need to be moved from Betanet to Mainnet
+* Handles are for display and discovery purposes only
 
 ## Discovery
 
 Contract addresses are not easy to remember and while some networks have had success with just numbers,
 it is most common to think of text based handles for finding other people on a network.
+DSNP handles are an easy way to allow easy user lookup.
 
-## Graph Connection Ids
+## Homograph Attack Mitigation
+
+UTF-8 support for handles opens handle users up to [homograph attacks](https://en.wikipedia.org/wiki/IDN_homograph_attack), not to mention case-sensitivity issues.
+This issue is of ongoing discussion both for the DSNP as well as in for ICANN domain names and other projects working with internationalization support.
+
+Because the DSNP Id is stable, attacks would only be successful in cases where the DSNP Id were unknown.
+[Punycode](https://en.wikipedia.org/wiki/Punycode) would resolve some issues, but would require losing true internationalization.
+
+### Current Mitigation Strategies
+
+Clients resolving handles MUST implement a method to detect potential homographs and check both user settings and potentially check the registry for additional potential matching DSNP Ids.
+
+## DSNP Ids
 
 Ethereum contract addresses are currently 160 bit values which is much larger than needed for unique identification.
 Graph connections can be reduced to just 64 bit identifiers.
 
+**Remember:** Only DSNP Ids are safe for long term data connections.
+
+* DSNP Identity contracts are not guaranteed long term for upgrade and migration reasons
+* Handles are subject to homograph attacks
+
+
 ## EIP 721
 
 While the NFT standard is not necessary to support, it provides standard ownership transfer and management tools.
+
+TODO: Research the possibilities of EIP 721 support.
 
 ## Rejected Alternatives
 
@@ -43,25 +65,7 @@ While the NFT standard is not necessary to support, it provides standard ownersh
   * Overly complex and expensive for the simple use case of handles
   * Support can be added later through a custom ENS resolver
   * Didn't provide a numerical id for efficient graph storage
-
-
-## Discussion Points
-
-* Should the interface support reverse resolutions? (id -> handle, address -> id/handle)
-  * YES
-* Should we still match the ENS Resolver spec [EIP 137](https://eips.ethereum.org/EIPS/eip-137#resolver-specification)?
-  * NO, stay away so that the future can implement it if wanted
-* Should we name this registry or resolver or split?
-  * NO 
-* Do we need the uint64 id or should we just use the handle?
-  * YES
-* Are we going to hash the handle (ENS style) client side?
-  * NO, we will store the strings
-* Can more than one handle point to the same social identity?
-  * NO
-* Are we case sensitive / normalizing?
-  * Strings are UTF-8
-  * Add a new story to investigate and a note that it is an open question
+  * Still subject to homograph attacks
 
 ## Contract Interface
 
@@ -74,7 +78,7 @@ interface IRegistry {
     /**
      * @dev Log when a new handle is registered
      * @param addr The address the handle is pointing at
-     * @param id The numerical id for use with the graph
+     * @param id The numerical DSNP Id for storing references
      * @param handle The actual UTF-8 string used for the handle 
      */
     event DSNPHandle(address addr, uint64 id, string handle);
@@ -82,7 +86,7 @@ interface IRegistry {
     /**
      * @dev Log when a the handle address is changed
      * @param addr The new address the handle is pointing at
-     * @param id The numerical id for reference 
+     * @param id The numerical DSNP Id for reference 
      */
     event DSNPHandleChange(address addr, uint64 id);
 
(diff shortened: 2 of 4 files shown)


pages/Identity/Factory.md
```diff
@@ -0,0 +1,216 @@
+---
+name: Factory
+route: /Identity/Factory
+menu: Factory
+---
+
+# Identity Factory
+
+The lest expensive way to create a new identity is through a identity factory.
```
shannonwells, line 9:
```suggestion
The least expensive way to create a new identity is through a identity factory.
```

pages/Identity/Factory.md
```diff
@@ -0,0 +1,216 @@
...
+## Specification Status
+
+| Version | Status |
+---------- | ---------
+| 0.1     | Draft |
+
+## Purpose
+1. Describe how an Identity Factory can create an identity
+1. Describe how an Identity Factory can allow someone else to pay for the creation an identity
+1. Constrict the creation of identities without owner permission
```
shannonwells, line 21: Restrict?

pages/Identity/Registry.md
```diff
@@ -0,0 +1,154 @@
...
+     * @dev Alter a handle resolution
+     * @param newAddr New address for the handle to point at
+     * @param handle The handle to alter
+     * 
+     * MUST be called by someone who is authorized on the contract via `EIP 1271(oldAddr).isValidSignature`
+     * MUST emit DSNPHandleOwnershipChange
+     */
+    function setAddr(address newAddr, string handle) external;
+
+    /**
```
shannonwells, line 141: This is a duplicate.

pages/Identity/Registry.md
```diff
@@ -0,0 +1,154 @@
...
+    function resolve(string handle) view external returns (address);
+
+    /**
+     * @dev Resolve a handle to a contract address
+     * @param id The id to resolve
+     * 
+     * @throws if not found
+     * @returns Address of the contract
+     */
+    function resolveById(uint64 id) view external returns (address);
```
shannonwells, line 97: maybe resolveToAddress?

pages/Identity/Registry.md
```diff
@@ -0,0 +1,154 @@
...
+     * @dev Resolve a handle to a contract address
+     * @param handle The handle to resolve
+     * 
+     * @throws if not found
+     * @returns Address of the contract
+     */
+    function resolve(string handle) view external returns (address);
+
+    /**
+     * @dev Resolve a handle to a contract address
```
shannonwells, line 91: The description here is the same as above

pages/Identity/Registry.md
```diff
@@ -0,0 +1,154 @@
...
+
+## Rejected Alternatives
+
+* Why not the Ethereum Name Service ([ENS](https://ens.domains/))?
+  * Overly complex and expensive for the simple use case of handles
+  * Support can be added later through a custom ENS resolver
+  * Didn't provide a numerical id for efficient graph storage
+
+
+## Discussion Points
```
wilwade, line 50: Discussion in a meeting:

* Should the interface support reverse resolutions? (id -> handle, address -> id/handle)
  * YES, it is just better to be able to reverse
* Should we still match the ENS Resolver spec [EIP 137](https://eips.ethereum.org/EIPS/eip-137#resolver-specification)?
  * NO, stay away so that the future can implement it if wanted
* Should we name this registry or resolver or split?
  * NO, keep it together as there is no need to follow the ENS style split
* Do we need the uint64 id or should we just use the handle?
  * uint64 would be really nice, let's try to keep it and adjust if the implementation makes it difficult.
* Are we going to hash the handle (ENS style) client side?
  * NO, we will store the strings (required for reverse resolutions)
* Can more than one handle point to the same social identity?
  * NO. While there are some ways around this, the registry does not allow for it. It would make name squatting easier and.
* Are we case sensitive / normalizing?
  * Strings are UTF-8
  * Add a new story to investigate and a note that it is an open question
* Do we need a way to have names expire or be garbage collected?
  * Maybe? Future discussions.
* Can we represent ownership of a name without having to have a contract behind that name?
  * Yes we could potentially allow for wallets to own a name, but that would not then resolve to a DSNP Identity, but it raises other questions:
    * Would we allow that key to sign messages?
    * Would we test against the interface to check?
    * Would this allow for too much flexibility?

pages/Identity/Registry.md
```diff
@@ -0,0 +1,154 @@
...
+     * @param r ECDSA Signature r value
+     * @param s ECDSA Signature s value
+     * @param v EIP-155 calculated Signature v value
+     * @param newAddr New address for the handle to point at
+     * @param handle The handle to alter
+     * 
+     * MUST be signed by someone who is authorized on the contract via `EIP 1271(oldAddr).isValidSignature`
+     * MUST emit DSNPHandle
+     */
+    function setAddrSignature(bytes32 r, bytes32 s, uint32 v, address newAddr, string handle) external;
```
wilwade, line 152: Discussion:
> Should the newAddr be required to "allow" the assignment (instead of just the oldAddr)?

The discussion resolved to say yes, but in some ways this conflicts with the choice of only allowing 1:1 name to identity address. Will discuss more.

wilwade (commented):
Notes from a meeting of the spec devs

pages/Identity/Factory.md
```diff
@@ -0,0 +1,216 @@
...
+    /**
+     * @dev Creates a new identity with the message sender as the owner
+     * @param logic The Logic address to use for identity creation
+     * 
+     * @returns The address of the newly created Identity
+     */
+    function createUpgradable(address logic) public returns (address);
+
+    /**
+     * @dev Creates a new identity with the erecover address as the owner
```
aneyzberg, line 139:
```suggestion
     * @dev Creates a new identity with the ecrecover address as the owner
```

pages/Identity/Factory.md
```diff
@@ -0,0 +1,216 @@
...
+    /**
+     * @dev Creates a new identity with the message sender as the owner
+     * @param beacon The beacon address to use for identity creation
+     * 
+     * @returns The address of the newly created Identity
+     */
+    function createBeacon(address beacon) public external returns (address);
+    
+    /**
+     * @dev Creates a new identity with the erecover address as the owner
```
aneyzberg, line 186:
```suggestion
     * @dev Creates a new identity with the ecrecover address as the owner
```

pages/Identity/Registry.md
```diff
@@ -0,0 +1,173 @@
...
+     * @param id The numerical DSNP Id for reference 
+     */
+    event DSNPHandleChange(address addr, uint64 id);
+
+    /**
+     * @dev Register a new handle
+     * @param addr Address for the handle to point at
+     * @param handle The handle for the address
+     * 
+     * MUST NOT allow a registration of an addr that is already attached to a different handle
```
wilwade, line 98: If handles are primarily for discovery, do we care if an identity delegation contract is attached to more than one handle?

Let's take the case where someone has two twitter accounts. They don't care that people know the same person is behind both accounts, merely that they have different "accounts" for different content.

This is different than the question of can two handles point to the SAME account.

It comes down to the question of "what is an account?". Is the account the identifier (the DSNP Id) or the ownership?

pages/Identity/Registry.md
```diff
@@ -0,0 +1,173 @@
...
+     * @dev Alter a handle resolution
+     * @param newAddr New address for the handle to point at
+     * @param handle The handle to alter
+     * 
+     * MUST NOT allow a registration of an newAddr that is already attached to a different handle
+     * MUST be called by someone who is authorized on the contract via `EIP 1271(oldAddr).isValidSignature`
+     * TODO: FIX THE ISSUE OF newAddr not being a part of the creation
+     * MUST emit DSNPHandleOwnershipChange
+     */
+    function setAddr(address newAddr, string handle) external;
```
wilwade, line 128: IF the handle is merely for discovery, should there be a way to change your handle?

Should handles be separate (more like an NFT)?

pages/Identity/Registry.md
```diff
@@ -0,0 +1,173 @@
...
+    function resolveHandleToId(string handle) view external returns (uint64);
+
+    /**
+     * @dev Resolve a DSNP Id to a contract address
+     * @param id The DSNP Id to resolve
+     * 
+     * @throws if not found
+     * @returns Address of the contract
+     */
+    function resolveIdToAddress(uint64 id) view external returns (address);
```
wilwade, line 171: We don't have a `resolveAddressToId` for reverse lookup, but do we want that?

pages/Identity/Registry.md
wilwade: Still need to choose if the newAddr needs to allow the assignment, however the 1:1 name to contract address require was relaxed as it was irrelevant as one could just set all the delegations the same on different contracts anyway.

pages/Identity/Registry.md
wilwade: The thinking of the DSNP Id as the "account" and the contract as merely ownership and other abilities means there is no reason to require two handle pointing at the same contract address (albeit suggested to keep separate)

Never however can more than one handle be pointed at the same DSNP Id.

pages/Identity/Registry.md
wilwade: Added a lot of details on how to do lookups in any which way someone might want.

pages/Identity/Registry.md
wilwade: Added a way to change handles.

pages/Identity/Registry.md
```diff
@@ -0,0 +1,154 @@
...
+
+Contract addresses are not easy to remember and while some networks have had success with just numbers,
+it is most common to think of text based handles for finding other people on a network.
+
+## Graph Connections
+
+Ethereum contract addresses are currently 160 bit values which is much larger than needed for unique identification.
+Graph connections can be reduced to just 64 bit identifiers.
+
+## EIP 721
```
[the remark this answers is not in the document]
wilwade: Final information is added into the spec, but basically 721 provides what would be for us a broken or duplicate ownership model. Perhaps something to investigate for Mainnet, but not Betanet.

pages/Identity/Registry.md
```diff
@@ -6,21 +6,228 @@ menu: Identity
...
+## Resolutions
+
+Resolutions are possible between any of the three pieces of data: Handle, DSNP Id, and Contract Address.
+While a utility method is provided for ease of moving from handle to the contract address,
+the other resolutions require using contract log events.
+
+### Current Handle -> Current Contract Address
+
+The `IRegistry.resolveHandleToAddress` method is the most efficient for discovery of current values.
+Perform a log search using the `DSNPId` event to discover all contract addresses and DSNP Ids that have had a given handle.
```
aneyzberg, line 74: this line and line 79 say the same thing. Are both of these functions returning DSNP ids and contract addresses. It seems like one should resolve to ID and one should resolve to contract address?

pages/Identity/Registry.md
```diff
@@ -6,21 +6,228 @@ menu: Identity
...
+    function registerByEIP712Sig(bytes32 r, bytes32 s, uint32 v, address addr, string handle) external returns (uint64);
+
+    /**
+     * @dev Alter a DSNP Id resolution address
+     * @param newAddr Original or new address to resolve to
+     * @param handle The handle to modify
+     * 
+     * MUST be called by someone who is authorized on the contract
+     *      via `IDelegation(oldAddr).isAuthorizedTo(oldAddr, Permission.OWNERSHIP_TRANSFER, block.number)`
+     * TODO: FIX THE ISSUE OF newAddr not being a part of the creation
```
aneyzberg, line 167: what do you mean here - with `newAddr` not being part of creation?

aneyzberg (approved):
LGTM just a couple small comments

shannonwells (changes requested):
A couple of questions, and there are a couple of places where I think it could be made a little more clear to readers who don't have all the context, or who aren't as familiar with the subject matter.

.spelling
```diff
@@ -81,3 +81,6 @@ NFT
 ENS
 uint64
 ecrecover
+Punycode
```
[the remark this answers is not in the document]
wilwade: 🙇

pages/Identity/Registry.md
```diff
@@ -6,21 +6,228 @@ menu: Identity
...
+ */
+interface IRegistry {
+
+    /**
+     * @dev Log when a resolution address is changed
+     * @param id The DSNP Id 
+     * @param addr The address the DSNP Id is pointing at
+     * @param handle The actual UTF-8 string used for the handle 
+     */
+    event DSNPId(uint64 indexed id, address indexed addr, string indexed handle);
```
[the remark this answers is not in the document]
wilwade: Riffing off your name, I wonder if `DSNPRegistryUpdate` is a better name. The Id is in there, but as it has all three pieces of information it is really a registry update now (it went through several previous names and at one time was three events!)

ID vs Id in the other comment thread.

pages/Identity/Registry.md
```diff
@@ -6,21 +6,228 @@ menu: Identity
 
 # Identity Registry
 
+A registry allows for distinct user identifiers and human readable discovery of network members.
+The DSNP Id Registry is a simple contract that allows switching handles, identity contracts, and chain migration
```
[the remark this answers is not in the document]
wilwade: (Goes down a bit of a rabbit trail of Id vs ID. Hehehe or it could be the DSNP id or DSNP super ego)

I was thinking of it as an abbreviation of `DSNP Identifier` instead of `DSNP Identity Document`, but I don't think I have a strong preference ether way. When we use it as a variable in the interface we just use `id`, so no real guidance there. In other code I would expect either `dsnpId` or `dsnpID` to be fine, although likely the first in case additional words come after it (for readability: `dsnpIdForSending` is easier to read imho than `dsnpIDForSending` although either would be better with a serif on that capital I.)

pages/Identity/Registry.md
```diff
@@ -6,21 +6,228 @@ menu: Identity
...
 
 ## Assumptions
 
-TBD
+* Ids will need to be moved from Betanet to Mainnet
+* Handles are for display and discovery purposes only
+
+## Discovery via DSNP Handles
+
+Contract addresses or numerical ids are not easy to remember.
```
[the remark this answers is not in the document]
wilwade: The lines were just to follow the pattern of one line per sentence, but I think you are right bullet points is how I wrote them.

pages/Identity/Registry.md
```diff
@@ -6,21 +6,228 @@ menu: Identity
...
+A user MAY NOT register multiple handles that point to the same DSNP Id.
+Handles must be unique.
+
+### Homograph Attack Mitigation
+
+UTF-8 support for handles opens handle users up to [homograph attacks](https://en.wikipedia.org/wiki/IDN_homograph_attack), not to mention case-sensitivity issues.
+This issue is of ongoing discussion both for the DSNP as well as in for ICANN domain names and other projects working with internationalization support.
+
+Because the DSNP Id is stable, attacks would only be successful in cases where the DSNP Id were unknown.
+[Punycode](https://en.wikipedia.org/wiki/Punycode) would resolve some issues, but would require losing true internationalization.
```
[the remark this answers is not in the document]
wilwade: Cleaned up:

> [Punycode](https://en.wikipedia.org/wiki/Punycode) is used by some software to prevent homographs by encoding all non-Latin characters into Latin characters.
So punycode does not properly present non-Latin characters which isn't reaching the level of internationalization support desired by the DSNP.

pages/Identity/Registry.md
```diff
@@ -6,21 +6,228 @@ menu: Identity
...
+The most recent event (the one with the highest block number), will give the current handle and contract address for the given DSNP Id.
+
+## EIP 721
+
+DSNP support for the NFT standard [EIP 721](https://eips.ethereum.org/EIPS/eip-721) for handles is not available for [Betanet](/Networks).
+It will be reconsidered for Mainnet.
+
+### NFT Concerns
+
+* EIP 721 ownership standards don't need permissioned levels
```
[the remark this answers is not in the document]
wilwade: > * The EIP 721 standard has its own ownership and permission system that is too limited for use across the DSNP.
> * Supporting two ownership systems adds unneeded complexity

pages/Identity/Registry.md
```diff
@@ -6,21 +6,228 @@ menu: Identity
...
+* Dual ownership standard support adds unneeded complexity
+* Identity contract ownership requires receive and transfer support
```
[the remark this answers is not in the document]
wilwade: > * Identity contract ownership would require additional complexity to receive and transfer 721 tokens

This version just assumes that complexity is bad, but it is listed under concerns.

pages/Identity/Registry.md
wilwade: That isn't super clean. How does this sound:

> To retrieve historical values, perform a log search using the `DSNPRegistryUpdate` event.

Or we could just remove them. The section "Other Lookups & Historical Values" has the information. The key is that that event has all three pieces of data, so it can be used to perform lookups in any direction.

pages/Identity/Registry.md
```diff
@@ -6,21 +6,228 @@ menu: Identity
...
+     * @param r ECDSA Signature r value
+     * @param s ECDSA Signature s value
+     * @param v EIP-155 calculated Signature v value
+     * @param newAddr New address for the DSNP Id to point at
+     * @param handle The handle to modify
+     * 
+     * MUST be signed by someone who is authorized on the contract
+     *      via `IDelegation(oldAddr).isAuthorizedTo(ecrecovedAddr, Permission.OWNERSHIP_TRANSFER, block.number)`
+     * MUST check that newAddr implements IDelegation interface   
+     * TODO: FIX THE ISSUE OF newAddr not being a part of the creation
```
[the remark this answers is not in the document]
wilwade: See https://github.com/LibertyDSNP/spec/pull/30#discussion_r619191842

pages/Identity/Registry.md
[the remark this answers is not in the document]
rosalinekarr: I like "Id" a little better, but it's not a strong opinion. We could just side-step the issue here by writing out "Identifier."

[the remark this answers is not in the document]
shannonwells: >I was thinking of it as an abbreviation of DSNP Identifier instead of DSNP Identity Document
Yea that makes sense.

TBH (or Tbh? if you prefer) I like dsnpAndThenTheCamelCase but ¯\_(ツ)_/¯

pages/Identity/Registry.md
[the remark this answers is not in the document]
shannonwells: DSNPRegistryUpdate works too

shannonwells (approved):

pages/Identity/Registry.md
aneyzberg: oh I see. got it. So then is this still a TODO?

pages/Identity/Registry.md
aneyzberg: Yes, I think what you suggested sounds better

enddynayn (approved):

pages/Identity/Registry.md
wilwade: @aneyzberg I think it is something we want to consider if we should do anything about it. I added it as an issue: https://github.com/LibertyDSNP/spec/issues/33

rosalinekarr (approved):

PR #31, opened 2021-04-19
Features DSNP Message Signatures

rosalinekarr:
Problem
=======
We need to define the signature generation specification for DSNP messages so messages can be authenticated in a reliable and predictable way.

[#177252383](https://www.pivotaltracker.com/story/show/177252383)
[#177681451](https://www.pivotaltracker.com/story/show/177681451)
[#177164341](https://www.pivotaltracker.com/story/show/177164341)

Solution
========
Define the signature spec.


pages/Messages/Signatures.md
```diff
@@ -0,0 +1,44 @@
...
+
+## Signing Messages
+
+The process for generating signatures for DSNP messages consists of three distinct steps:
+
+1. Serialize the message
+1. Hash the serialized string
+1. Sign the hash
+
+Before serializing, keys in the DSNP message should be sorted alphabetically using [default JavaScript string comparison](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Collator) order.
```
shannonwells, line 34: I thought we were getting rid of this requirement and accepting that there may be different hashes for the same message.

shannonwells (changes requested):
I was pretty sure we got rid of the key ordering part.  Did that change/

pages/Messages/Signatures.md
rosalinekarr: We removed the requirement for the hash of the activity pub content, since we don't control the keys in an activity pub object and sorting them becomes complex, but we still need this requirement for the hash of the DSNP message to guarantee we always get a consistent result.

Plus, it shouldn't be a problem for the DSNP object, since we control the all the keys in it and can guarantee they're always ANSI characters, which we can sort.

aneyzberg (approved):
this looks good to me

shannonwells (approved):

enddynayn (approved):

PR #37, opened 2021-04-29
Features/message identifier 177873013

rosalinekarr:
Problem
=======
We need a way to uniquely identify DSNP messages posted by different users.
[#177873013](https://www.pivotaltracker.com/story/show/177873013)

Solution
========
Added a spec file for Message Identifiers

### DIFF
--- a/pages/Messages/Identifiers.md
+++ b/pages/Messages/Identifiers.md
@@ -35,11 +35,8 @@ In this example, `dsnp://` is the protocol indicating that the latest version of
 
 ### Protocol
 
-The protocol section of an identifier can be used to indicate which version of the DSNP specification should be used to identify content.
-If present, any number after the string `dsnp` should indicate the specific version.
-For example, `dsnp2://` would indicate that version 2 of the specification should be used.
-
-If no version number is provided at all, i.e. `dsnp://`, then the most recent available version of the spec should be used.
+As of this specification version, the protocol string should always be the string `dsnp://`.
+In future specification versions, this string may change to indicate non-backward-compatible changes.
 
 ### User Identifier
 
--- a/pages/Messages/Identifiers.md
+++ b/pages/Messages/Identifiers.md
@@ -1,0 +1,91 @@
+---
+name: Identifiers
+route: /Messages/Identifiers
+menu: Messages
+---
+
+# Messages Identifiers
+
+## Specification Status
+
+| Version | Status |
+---------- | ---------
+| 0.1     | Tentative |
+
+## Purpose
+
+1. Describe a means of uniquely identifying DSNP content messages
+1. Describe potential extensions of identifiers for the future
+
+## Assumptions
+
+* All assumptions from [DSNP Messages](/Messages/Overview)
+
+## Identifiers
+
+DSNP message identifiers at minimum consist of a string with three parts, a protocol, a user identifier and a content hash, like so:
+
+```
+dsnp://0123/456789abcdef0123
+```
+
+In this example, `dsnp://` is the protocol indicating that the latest version of this specification should be used.
+`0123` is the identifier representing a user as described in the [Identity Registry](/Identity/Registry) specification.
+`456789abcdef0123` is the hash of the content as described in the [Messages Serialization](/Messages/Serialization) specification.
+
+### Protocol
+
+The protocol section of an identifier can be used to indicate which version of the DSNP specification should be used to identify content.
+If present, any number after the string `dsnp` should indicate the specific version.
+For example, `dsnp1.2.1://` would indicate that version 1, subversion 2, patch release 1 of the specification should be used.
+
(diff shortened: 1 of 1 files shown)


pages/Messages/Identifiers.md
```diff
@@ -0,0 +1,91 @@
...
+If the patch release or subversion number is omitted, the latest available subversion and patch release should be used.
+For example, `dsnp2://` would indicate the most up to date subversion and patch release under version 2 of the DSNP should be used.
+
+If no version number is provided at all, i.e. `dsnp://`, then the most recent available version of the spec should be used.
+
+### User Identifier
+
+The user identifier section of the identifier string should adhere to the relevant version of the [Identity Registry](/Identity/Registry) specification.
+Generally, this will be a hexadecimal representation of the user's 64 bit registry identifier number.
+For example, `2B02` would indicate user number 690.
```
rosalinekarr, line 51: I really wanted to throw in a Kurt Vonnegut reference here, but unfortunately, [2BR02B](https://en.wikipedia.org/wiki/2_B_R_0_2_B) is more than 64 bits in hex.

pages/Messages/Identifiers.md
```diff
@@ -0,0 +1,91 @@
...
+
+If the patch release or subversion number is omitted, the latest available subversion and patch release should be used.
+For example, `dsnp2://` would indicate the most up to date subversion and patch release under version 2 of the DSNP should be used.
+
+If no version number is provided at all, i.e. `dsnp://`, then the most recent available version of the spec should be used.
+
+### User Identifier
+
+The user identifier section of the identifier string should adhere to the relevant version of the [Identity Registry](/Identity/Registry) specification.
+Generally, this will be a hexadecimal representation of the user's 64 bit registry identifier number.
```
wilwade, line 50: I think we should make this more firm for now.

```suggestion
This MUST be a hexadecimal representation of the user's 64 bit registry identifier number.
```

pages/Messages/Identifiers.md
```diff
@@ -0,0 +1,91 @@
...
+
+In this example, `dsnp://` is the protocol indicating that the latest version of this specification should be used.
+`0123456789abcdef` is the identifier representing a user as described in the [Identity Registry](/Identity/Registry) specification.
+`0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef` is the hash of the content as described in the [Messages Serialization](/Messages/Serialization) specification.
+
+### Protocol
+
+The protocol section of an identifier can be used to indicate which version of the DSNP specification should be used to identify content.
+If present, any number after the string `dsnp` should indicate the specific version.
+For example, `dsnp1.2.1://` would indicate that version 1, subversion 2, patch release 1 of the specification should be used.
```
wilwade, line 40: Hmm... I wonder if we want to allow this level of detail. Especially as this identifier is going to be stored in the announcement (inReplyTo), it means that we cannot go back and change it.

I think we should just lock the version and only bump for breaking changes aka: `dsnp://` -> `dsnp2://` and require it be that way on all v2.

A non-breaking change would be like supporting `@username` instead of the DSNP Id. (a breaking change would be requiring 0x prefixed hex values).

pages/Messages/Identifiers.md
```diff
@@ -0,0 +1,91 @@
...
+
+The content hash section of the identifier string should be a [keccak-256](https://en.wikipedia.org/wiki/SHA-3) hash of the content of the message as stored at the URI provided in the DSNP announcement.
+This string should be identical to the `contentHash` field as described in the [Message Overview](/Messages/Overview).
+
+## Optional Features
+
+In addition to the specification above, several extensions of the identifier string may optionally be supported, however support is not yet guaranteed across all implementations.
+Each of these extensions should be backward compatible in that clients may ignore the added functionality and treat the provided string like a basic message identifier.
+
+### Alternative Formatting
```
wilwade, line 63: Re: formatting and request options

I like these, but I wonder if they are in the wrong place?

We have the identifier piece (aka `inReplyTo`) which shouldn't ever have any of this extensions. However I really do like this request bit for services offering easy access to content (and indexer for example).

Thoughts on how to separate out the two use cases?

Perhaps we need a name for the "raw"|"base"|etc... dsnp uri?

I can see going ahead and supporting the `@username` for the request case as well.

Might also want to pull in @pfrank13 to this discussion.

wilwade (dismissed):
Two potential changes and one big discussion point

pages/Messages/Identifiers.md
rosalinekarr: Fixed

pages/Messages/Identifiers.md
rosalinekarr: I'll omit this section for now and we can circle back on the idea later.

wilwade (approved):
🚅

wilwade (approved):

enddynayn (approved):
easy to follow 👍

PR #40, opened 2021-05-03
remove authorization from initial registration

### LINKED ISSUE #33: Other user can assign ownership of a DSNP Id and handle to my contract without my permission
### Steps to reproduce
1. Register a new DSNP Id with a handle and point it at an identity contract owned by user A.
2. Transfer ownership of that DSNP Id to a contract owned by user B

### Expected behavior

User A AND user B should be required to approve the transfer of the DSNP Id to user B

### Actual behavior

Only user A is required to approve the transfer of the DSNP Id

### Any other comments?

It is possible this is the desired outcome and it is an ok result.

acruikshank:
closes #33 

Problem
=======
Authorization for a new identity from its delegation contract isn't necessary. It can be circumvented by `changeAddress` if it does not authenticate against the new contract, and transferring between delegation contracts is complicated if it does. 
[link to Pivotal Tracker #178010820](https://www.pivotaltracker.com/story/show/178010820)

Solution
========
This PR replaces the requirement that `register` and `changeAddress` authorize against the contract with the requirement that it confirm the contract implements IDelegation. It also removes the `registerByEIP712Sig` since the `register` method now does the same thing and is more cost effective.

Change summary:
---------------
* Remove language about call to `IDelegation.isAuthorizedTo` in `register`.
* Add language about call to EIP 165 `supportsInterface` in `register`.
* Remove `registerByEIP712Sig` method.
* Add language about call to EIP 165 `supportsInterface` in `changeAddress`.

Steps to Verify:
----------------
1. Read the diff

### DIFF
--- a/pages/Identity/Registry.md
+++ b/pages/Identity/Registry.md
@@ -14,7 +14,7 @@ while maintaining all graph connections, public and private.
 
 | Version | Status |
 ---------- | ---------
-| 0.1     | Tentative |
+| 0.2     | Tentative |
 
 ## Purpose
 1. Describes how the Identity Registry resolves a DSNP Id to an identity contract address
--- a/pages/Identity/Registry.md
+++ b/pages/Identity/Registry.md
@@ -137,26 +137,10 @@ interface IRegistry {
      * @param handle The handle for discovery
      * 
      * MUST reject if the handle is already in use
-     * MUST be called by someone who is authorized on the contract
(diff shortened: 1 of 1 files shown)


wilwade (changes requested):
One change: We'll want to bump the version number at the top of the file.

wilwade (approved):

PR #41, opened 2021-05-04
add nonce parameter to EIP 712 signatures with justification

### LINKED ISSUE #38: Registry EIP 712 methods as written are subject to replay attacks
### Problem

The EIP 712 methods as written take only the signature information and the parameters to be changed. Once one of these methods is called, it's parameters become public and anyone can call the same method with the same signature information to act on behalf of the original sender. This means the registry cannot defend against the following replay attacks:

1. If `registerByEIP712Sig` is used to register a handle and then the handle is change (via `changeHandle` or `changeHandleByEIP712Sig` to free the handle. Anyone may re-register the handle to the same contract using the original `registerByEIP712Sig` method. Note that this becomes irrelevant if the resolution to #33 is to remove authorization on registration.
2. If `changeAddressByEIP712Sig` is used to change the contract address once and then any method is used to change the contract again, an attacker can use the original method call to change the address back. This is of particular concern if the point of the second change was to move the registration away from a contract that is no longer in the user's control.
3. If `changeHandleByEIP712Sig` is used to change the handle for an identity, any subsequent handle change can be rolled back by anyone using the original method.

Since the contract has no way to distinguish a legitimate EIP 712 request to undo some action from a replay attack, the spec needs some modification to address these issues.

The simplest solution is to require a nonce as an additional parameter to each method. The nonce will need to be stored with the registration. It will also need to be included within the signed data, and passed as a parameter to the EIP 712 methods. So legitimate EIP 712 requests will have a new nonce, and the old nonce in replay attacks can be rejected.

acruikshank:
closes #38

Problem
=======
As written, the EIP 712 methods in Registry have no way to guard against replay attacks.
[#178007793](https://www.pivotaltracker.com/story/show/178007793)

Solution
========
Introduce a nonce that must be stored along with registration issues, and then require the nonce parameter to be included in the signature and passed as a parameter to the EIP 712 methods. This will allow the registry to reject transactions that have already been played.

Change summary:
---------------
* Add section about EIP 712, replay attacks, and the nonce
* Add nonce parameter to EIP 712 methods
* Add view method to retrieve nonce.

### DIFF
--- a/pages/Identity/Registry.md
+++ b/pages/Identity/Registry.md
@@ -14,11 +14,7 @@ while maintaining all graph connections, public and private.
 
 | Version | Status |
 ---------- | ---------
-<<<<<<< HEAD
-| 0.2     | Tentative |
-=======
 | 0.3     | Tentative |
->>>>>>> add nonce parameter to EIP 712 signatures with justification
 
 ## Purpose
 1. Describes how the Identity Registry resolves a DSNP Id to an identity contract address
@@ -59,7 +55,11 @@ Clients resolving handles MUST implement a method to detect potential homographs
 
 ### EIP 712 Methods, Replay Attacks and Nonces
 
-The Registry supports EIP 712 methods to permit a second party to pay gas costs for address and handle changes. Once an EIP 712 transaction is made, anyone may replay that action without further authorization. This breaks our security guarantees when the registration owner has made an additional change to either address or handle. To mitigate this, the Registry contract MUST store a nonce for every registration. When it receives an EIP 712 transaction, it MUST check that the nonce parameter matches the stored nonce, and it MUST increment the stored nonce if the transaction succeeds.
+The Registry supports EIP 712 methods to permit a second party to pay gas costs for address and handle changes.
+Once an EIP 712 transaction is made, anyone may replay that action without further authorization.
+This breaks our security guarantees when the registration owner has made an additional change to either address or handle.
+To mitigate this, the Registry contract MUST store a nonce for every registration.
+When it receives an EIP 712 transaction, it MUST check that the nonce parameter matches the stored nonce, and it MUST increment the stored nonce if the transaction succeeds.
 
 If a handle is changed, the registry MUST preserve the stored nonce for the old handle.
 
@@ -183,7 +183,7 @@ interface IRegistry {
      * TODO: FIX THE ISSUE OF newAddr not being a part of the creation
      * MUST emit DSNPRegistryUpdate
      */
-    function changeAddressByEIP712Sig(bytes32 r, bytes32 s, uint32 v, address newAddr, string handle, uint64 nonce) external;
+    function changeAddressByEIP712Sig(bytes32 r, bytes32 s, uint32 v, address newAddr, string handle, uint32 nonce) external;
 
     /**
      * @dev Alter a DSNP Id handle
@@ -211,7 +211,7 @@ interface IRegistry {
      *      via `IDelegation(handle -> addr).isAuthorizedTo(ecrecovedAddr, Permission.OWNERSHIP_TRANSFER, block.number)`
      * MUST emit DSNPRegistryUpdate
      */
-    function changeHandleByEIP712Sig(bytes32 r, bytes32 s, uint32 v, string oldHandle, string newHandle, uint64 nonce) external;
+    function changeHandleByEIP712Sig(bytes32 r, bytes32 s, uint32 v, string oldHandle, string newHandle, uint32 nonce) external;
 
     /**
      * @dev Resolve a handle to a contract address
@@ -238,6 +238,6 @@ interface IRegistry {
      * @throws if not found
      * @returns expected nonce for next EIP 712 update
      */
-    function resolveHandleToNonce(string handle) view external returns (uint64);
+    function resolveHandleToNonce(string handle) view external returns (uint32);
 }
 ```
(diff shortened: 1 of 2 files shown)


pages/Identity/Registry.md
```diff
@@ -53,6 +53,10 @@ So punycode does not properly present non-Latin characters which isn't reaching
 
 Clients resolving handles MUST implement a method to detect potential homographs and check both user settings and potentially check the registry for additional potential matching DSNP Ids.
 
+### EIP 712 Methods, Replay Attacks and Nonces
+
+The Registry supports EIP 712 methods to permit a second party to pay gas costs for address and handle changes. Once an EIP 712 transaction is made, anyone may replay that action without further authorization. This breaks our security guarantees when the registration owner has made an additional change to either address or handle. To mitigate this, the Registry contract MUST store a nonce for every registration. When it receives an EIP 712 transaction, it MUST check that the nonce parameter matches the stored nonce, and it MUST increment the stored nonce if the transaction succeeds.
```
enddynayn, line 58: ❤️

enddynayn (approved):

pages/Identity/Registry.md
```diff
@@ -161,14 +171,15 @@ interface IRegistry {
...
+     * @param nonce The expected nonce for this update (must match return value of resolveHandleToNonce())
      * 
      * MUST be signed by someone who is authorized on the contract
      *      via `IDelegation(oldAddr).isAuthorizedTo(ecrecovedAddr, Permission.OWNERSHIP_TRANSFER, block.number)`
      * MUST check that newAddr implements IDelegation interface
      * TODO: FIX THE ISSUE OF newAddr not being a part of the creation
      * MUST emit DSNPRegistryUpdate
      */
-    function changeAddressByEIP712Sig(bytes32 r, bytes32 s, uint32 v, address newAddr, string handle) external;
+    function changeAddressByEIP712Sig(bytes32 r, bytes32 s, uint32 v, address newAddr, string handle, uint64 nonce) external;
```
wilwade, line 182: @acruikshank Are we sure we want a 64 bit nonce? (address + id + nonce) 160+64+64 > 256 So our struct would take more than one storage slot. A 32 bit nonce would fit us into a single 256 slot 160+64+32.

wilwade (changes requested):
One option to consider for going from 64 bit nonces to 32 bit

pages/Identity/Registry.md
```diff
@@ -79,6 +85,10 @@ To retrieve historical values, perform a log search using the `DSNPRegistryUpdat
 The `IRegistry.resolveHandleToId` method is the most efficient for discovery of current values.
 To retrieve historical values, perform a log search using the `DSNPRegistryUpdate` event.
 
+### Current Handle -> Nonce
+
+The `IRegistry.resolveHandleToNonce` method is the only method for discovery of the current nonce.
```
wilwade, line 90: For me at least "current" meant most recently used (aka ethereum's getTransactionCount), but that is not what it means right?
```suggestion
The `IRegistry.resolveHandleToNonce` method is the only method for discovery of the next nonce to use.
```

pages/Identity/Registry.md
```diff
@@ -14,7 +14,11 @@ while maintaining all graph connections, public and private.
 
 | Version | Status |
 ---------- | ---------
+<<<<<<< HEAD
 | 0.2     | Tentative |
+=======
+| 0.3     | Tentative |
+>>>>>>> add nonce parameter to EIP 712 signatures with justification
```
rosalinekarr, line 21: Looks like some merge slipped in

pages/Identity/Registry.md
```diff
@@ -53,6 +57,12 @@ So punycode does not properly present non-Latin characters which isn't reaching
 
 Clients resolving handles MUST implement a method to detect potential homographs and check both user settings and potentially check the registry for additional potential matching DSNP Ids.
 
+### EIP 712 Methods, Replay Attacks and Nonces
+
+The Registry supports EIP 712 methods to permit a second party to pay gas costs for address and handle changes. Once an EIP 712 transaction is made, anyone may replay that action without further authorization. This breaks our security guarantees when the registration owner has made an additional change to either address or handle. To mitigate this, the Registry contract MUST store a nonce for every registration. When it receives an EIP 712 transaction, it MUST check that the nonce parameter matches the stored nonce, and it MUST increment the stored nonce if the transaction succeeds.
```
rosalinekarr, line 62: Could you break out each sentence on it's own line? As long as it's only a single newline, it won't show up on the rendered page, but it makes diffs easier to read.

rosalinekarr (changes requested):

pages/Identity/Registry.md
acruikshank: The current stored nonce should be next nonce to use (i.e. stored nonces should start at 0 and the first EIP 712 transaction should have a nonce of zero). I suppose the suggestion makes this a little more clear.

pages/Identity/Registry.md
acruikshank: I'll change it. I thought this would be handled by the number encoding when the struct is packed, but I can't find a definitive answer on that.

rosalinekarr (approved):

rosalinekarr:
looks like the spellchecker needs some words whitelisted

enddynayn (approved):

wilwade (approved):

PR #42, opened 2021-05-05
encapsulate EIP 712 method params to disambiguate serialization

acruikshank:
Problem
=======
The EIP 712 signature specification describes how to serialize structured data in a consistent and safe manner for signing. It only describes how to do this for transactions, byte strings, and a single (possibly nested) data structure. We are currently passing in 3 primitive parameters to each of our EIP712 methods.

We may implicitly think of these these parameters as forming the members of a single data structure, and sign it that way. EIP 712 requires that we create a type hash of this data structure with a name and hash it along with the rest of the data. With no actual data structure to reference, the name is ambiguous. Even the contents are a little ambiguous (all the parameters that come after the first 3 signature parameters).

**Additionally** the `v` parameter of the signature should be a uint8 instead of a uint32.

Solution
========
* Create an `AddressChange` and `HandleChange` struct to be passed as parameters to corresponding methods. This removes ambiguity around what exactly should be signed.
* Change signature `v` parameter to `uint8` and reorder signature parameters to match order of `erecover` signature to avoid confusion.

### DIFF
--- a/pages/Identity/Registry.md
+++ b/pages/Identity/Registry.md
@@ -181,9 +181,9 @@ interface IRegistry {
 
     /**
      * @dev Alter a DSNP Id resolution address by EIP-712 Signature
+     * @param v EIP-155 calculated Signature v value
      * @param r ECDSA Signature r value
      * @param s ECDSA Signature s value
-     * @param v EIP-155 calculated Signature v value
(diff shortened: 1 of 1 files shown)


aneyzberg (approved):
lgtm

PR #43, opened 2021-05-06
Spec/serialization update 177939555

rosalinekarr:
Problem
=======
Given our solution for generating batch files, we no longer need to serialize DSNP data except for generating hashes and signatures
[#177939555](https://www.pivotaltracker.com/story/show/177939555)

Solution
========
Update the spec accordingly


aneyzberg (approved):
lgtm

pages/Messages/Signatures.md
```diff
@@ -27,16 +27,19 @@ Optionally, anonymous messages may be provided with a zero hash in place of the
...
-Serialization should be done as described in the [Serialization spec](/Identity/Serialization).
-Next, the serialized string should be hashed using [keccak-256](https://en.wikipedia.org/wiki/SHA-3).
-Finally, the fixed length hash generated can then be signed using [secp256k1](https://google.com/search?hl=en&q=secp256k1) and the publishing user's private key.
+For encrypted messages, the encrypted bytes of the `dsnpData` field can be used directly without any serialization.
 
-Once the signature is generated, it should be provided along with the message contents to the announcer for inclusion in the next available batch.
+For non-encrypted messages, keys in the `dsnpData` object should be sorted alphabetically using [default JavaScript string comparison](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Collator) order to guarantee a single, unique string for any given object.
+Once sorted, the object should be serialized using the standard [stringify method](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify) available in NodeJS, most browsers or a variety of widely available libraries.
```
acruikshank, line 37: 1. This is a little unclear. The object itself can't be sorted by keys, so I'm not sure what's being stringified. Are nested objects also sorted by key? Does this require a custom stringification implementation? The ECMA spec stringify method permits a `replacer` function for numbers. I'm not sure how consistent double64 serialization is across JS implementations. I know default date serialization is not consistent. Are dates permitted? What JS types are and are not permitted in serialization?
2. Relying on Javascript serialization will make signature verification extremely difficult on chain (or really in any other language). I don't particularly like it, but EIP 712 serialization should at least be considered. Are there other language-neutral serialization specifications we could borrow?

pages/Messages/Signatures.md
```diff
@@ -27,16 +27,19 @@ Optionally, anonymous messages may be provided with a zero hash in place of the
...
 1. Sign the hash
 
-Before serializing, keys in the DSNP object should be sorted alphabetically using [default JavaScript string comparison](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Collator) order.
-Serialization should be done as described in the [Serialization spec](/Identity/Serialization).
-Next, the serialized string should be hashed using [keccak-256](https://en.wikipedia.org/wiki/SHA-3).
-Finally, the fixed length hash generated can then be signed using [secp256k1](https://google.com/search?hl=en&q=secp256k1) and the publishing user's private key.
+For encrypted messages, the encrypted bytes of the `dsnpData` field can be used directly without any serialization.
 
-Once the signature is generated, it should be provided along with the message contents to the announcer for inclusion in the next available batch.
+For non-encrypted messages, keys in the `dsnpData` object should be sorted alphabetically using [default JavaScript string comparison](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Collator) order to guarantee a single, unique string for any given object.
```
acruikshank, line 36: This references `Intl.Collator` which permits customization by language. The MDN page suggests this defaults to the default locale, which will cause inconsistent serializations. Perhaps this should specify the language as well (which I think would have to be `en_US`).

pages/Messages/Signatures.md
```diff
@@ -27,16 +27,19 @@ Optionally, anonymous messages may be provided with a zero hash in place of the
 
 The process for generating signatures for DSNP announcements consists of three distinct steps:
 
-1. Serialize the DSNP announcement
```
enddynayn, line 30: It would be nice if we provided an example. Can we provide an example?

pages/Messages/Signatures.md
acruikshank: I didn't realize that JS object order has been deterministic since es 2015, so that clarifies my question about how the object can be sorted and still use stringify. It might be better to point to https://tc39.es/ecma262/#sec-json.stringify than the MDN docs.

shannonwells (commented):
I don't have anything to add to what's already been mentioned.

pages/Messages/Signatures.md
rosalinekarr: Examples added

pages/Messages/Overview.md
```diff
@@ -82,7 +82,7 @@ This is what would be posted as a Log event in Ethereum:
 |-------|-------------|------|
```
shannonwells, line 82: Since we're not emitting individual messages, the entire Variable Announcement Format section can be deleted; it's not specific to serialization, so I can include it as part of my Batch spec changes instead if you like.

shannonwells (approved):

PR #44, opened 2021-05-06
ProxyCreated + Small fixes

wilwade:
Problem
=======
Transaction methods cannot have a return value, so we need to log an event for proxy creation:

> [Write methods] cannot return a result. If a result is required, it should be logged using a Solidity event (or EVM log), which can then be queried from the transaction receipt.
Re: https://docs.ethers.io/v5/api/contract/contract/#Contract--write

[Pivotal Tracker #12345678](https://www.pivotaltracker.com/story/show/177872358)

Solution
========
Add ProxyCreated event with the needed data

Change summary:
---------------
* Add ProxyCreated event with the needed data
* Update EIP 712 methods to match #42 
* Fix typo of @returns instead of @return

### DIFF
--- a/pages/Identity/Factory.md
+++ b/pages/Identity/Factory.md
@@ -16,7 +16,7 @@ Any contract that matches the [DSNP Identity](/Identity/Overview) interfaces is 
 
 | Version | Status |
 ---------- | ---------
-| 0.1     | Tentative |
+| 0.2     | Tentative |
 
 ## Purpose
 1. Describe how an Identity Factory can create an identity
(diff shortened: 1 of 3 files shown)


pages/Identity/Factory.md
```diff
@@ -75,26 +75,33 @@ Clones follow [EIP 1167](https://eips.ethereum.org/EIPS/eip-1167) for a **non-up
...
      * @dev [EIP 1167](https://eips.ethereum.org/EIPS/eip-1167) Proxy
+     * @param v EIP-155 calculated Signature v value
      * @param r ECDSA Signature r value
      * @param s ECDSA Signature s value
-     * @param v EIP-155 calculated Signature v value
      * @param logic The Logic address to use for identity creation
      * 
-     * @returns The address of the newly created Identity
+     * @dev This MUST emit ProxyCreated with the address of the new proxy contract
+     * @return The address of the newly created identity proxy contract
```
acruikshank, line 102: Is this still necessary? Seems like an added expense.

pages/Identity/Factory.md
```diff
@@ -118,36 +130,39 @@ interface IIdentityUpgradableFactory {
...
      */
     function getLogic() external view returns (address);
 
     /**
      * @dev Creates a new identity with the message sender as the owner
      *      and will be pointed at the default identity logic address.
      * 
-     * @returns The address of the newly created Identity
+     * @dev This MUST emit ProxyCreated with the address of the new proxy contract
+     * @return The address of the newly created identity proxy contract
```
acruikshank, line 142: same as above here and on down.

acruikshank (approved):
The 712 sig reorder and addition of the event looks good. Added a question about whether we still want to return a value from the write methods that now emit the same value? Is this for on-chain calls? Does it have an additional cost? If the answers are yes to the first or no to the second, I'm good.

pages/Identity/Factory.md
wilwade: It is an extra 236 gas, but it is needed in the case of other contracts calling it.

PR #45, opened 2021-05-11
Spec/message validation 177164337

rosalinekarr:
Problem
=======
We need a specification to validate DSNP message for inclusion in a batch, inclusion in an archive and display to users.
[#177164337](https://www.pivotaltracker.com/story/show/177164337)

Solution
========
Write the 'Message Validation' specification.

### DIFF
--- a/pages/Messages/Validation.md
+++ b/pages/Messages/Validation.md
@@ -137,22 +137,6 @@ The resulting hash would be `0x70ae98439569700ae8328f204ba496e4ac151dc117d08ac21
 Notice that this hash is generated including spaces and newlines exactly as they appear in the content body.
 The content body must be hashed byte for byte from the response of the URI with absolutely no processing.
 
-## Type Specific Validation
-
-Specific types of DSNP messages may also include their own specific validation checks depending on the semantic purpose of the message type.
-These specific validation checks follow:
-
-### Broadcasts
-
-1. All broadcast messages with a `"type"` field of `"Follow"` in the activity pub content must be preceded by a follow [Graph Change](/Messages/Overview#GraphChange) DSNP message with a followee matching the `"object"` field of the content.
-1. All broadcast messages with a `"type"` field of `"Like"` or `"Dislike"` in their activity pub content must be preceded by a [Reaction](/Messages/Overview#Reaction) DSNP message referring to the same object, i.e. that the `URI` message field of DSNP message identified by the `inReplyTo` field of the reaction is identical to the `"object"` field of the like or dislike activity pub content.
-1. All broadcast messages with a `"type"` field of `"Undo"` must include an `"object"` field with either a valid DSNP identifier referring to a previously announced broadcast message or a URI matching the URI of a previously announced broadcast event by the user.
-
-### Replies & Reactions
-
-1. All reply and reaction messages must include an `inReplyTo` field identifying an existing DSNP message.
-1. All reply and reaction messages must be from and signed by a user, the replier, for which the user who posted the original content referred to in the `inReplyTo` field has never previously published a broadcast message with an activity pub content `"type"` field of `"Block"` and `"object"` field containing the string `dsnp://` followed by the user id of the replier, thereby indicating that the original poster has previously blocked the replier.
-
 ## Optional Validations
 
 In addition to the aforementioned general validation and type specific validation checks, there are also a set of optional validation checks that implementers are encouraged to support, however end-users may choose to ignore them depending on their legal jurisdiction or personal preference.
--- a/pages/Messages/Validation.md
+++ b/pages/Messages/Validation.md
@@ -31,10 +31,6 @@ Announcers may choose to skip content validation checks in the interest of perfo
 Validation of announcement correctness will vary depending on which fields are present on a particular DSNP message type.
 Validating correctness for a message will consist of verifying that each field in the message is defined and meets the formatting rules listed in the subsections below for the given field type.
 
-### Addresses
-
-1. Address fields must be exactly 20 bytes.
-
 ### Emoji
 
 1. Emoji fields must not be empty.
@@ -73,6 +69,10 @@ Additionally, none of the following should be considered valid:
 1. URI fields must include meet all standards defined in [RFC3986](http://www.ietf.org/rfc/rfc3986.txt).
 1. URI fields must not refer to localhost or any reserved IP addresses as defined in [RFC6890](https://datatracker.ietf.org/doc/html/rfc6890)
 1. URI fields must use the `https` protocol. Support for other protocols may be added in the future.
+
+### User Ids
+
+1. User Id fields must be exactly 8 bytes.
 
 ## Announcement Authenticity
 
--- a/pages/Messages/Validation.md
+++ b/pages/Messages/Validation.md
@@ -136,36 +136,3 @@ For example, given a DSNP message with the following content at it's URI:
 The resulting hash would be `0x70ae98439569700ae8328f204ba496e4ac151dc117d08ac217daa15b412641f7`.
 Notice that this hash is generated including spaces and newlines exactly as they appear in the content body.
 The content body must be hashed byte for byte from the response of the URI with absolutely no processing.
-
-## Optional Validations
-
-In addition to the aforementioned general validation and type specific validation checks, there are also a set of optional validation checks that implementers are encouraged to support, however end-users may choose to ignore them depending on their legal jurisdiction or personal preference.
-As such, it is strongly encouraged to make these features configurable by the end-user in all implementations.
-
-### Copyright & Licensing
-
-Implementations may also choose or be required by law to invalidate content identified as violating certain laws, such as intellectual property, in the end-user's specific legal jurisdiction.
-These implementations may use content fingerprinting, shared lists or other methods to identify this content.
-This document will not define any standards for these methods except that before removing any potentially violating content implementers must verify that the offending content does not include proof of legal exemption, such as paid licensing, in the activity pub content of the message.
-
-These legal exceptions must be listed under an additional content field named `"licensing"` which may include either a URI pointing to the appropriate proof, an object to be defined by the enforcing legal body or an array consisting of multiple instances of either.
-For example, here is a sample activity pub object with fictional licensing:
-
-```json
-{
-  "@context": "https://www.w3.org/ns/activitystreams",
-  "type": "Video",
-  "name": "Excerpt of Yankees vs Nationals Game 2021",
-  "url": "http://sportsblog.com/yankeesnationals2021.mkv",
-  "duration": "5M",
-  "licensing": [
-    "https://mlb.com/licensing/sportsblogcom.json",
-    {
-      "authority": "ESPN",
-      "type": "paid",
-      "licensee": "SportsBlog.com",
-      "proof": "c3RyaW5naBc6453=w-9sSAw4naBc8533="
-    }
-  ]
-}
-```
--- a/pages/Messages/Validation.md
+++ b/pages/Messages/Validation.md
@@ -29,20 +29,7 @@ Announcers may choose to skip content validation checks in the interest of perfo
 ## Announcement Correctness
 
 Validation of announcement correctness will vary depending on which fields are present on a particular DSNP message type.
-In general, announcement correctness validation will consist of confirming that all necessary fields are present and that values in each field use a format appropriate for the field type.
-
-### Encrypted Messages
-
-Announcers who do not possess the necessary keys to decrypt an encrypted message cannot validate the message's correctness and must treat the message as valid so users with the appropriate keys may still access them in the produced batch.
-In contrast, clients and indexers who do not possess the necessary keys to decrypt an encrypted message should consider the message invalid and ignore it as it provides no value to the end-user.
-
-Optionally, announcers and indexers may choose to invalidate encrypted messages for which the content is obviously incorrect, i.e. the encrypted byte array isn't long enough to be a valid message, however no standard will be defined for this behavior and caution should be used to avoid incorrectly invalidating messages.
-
-If any validator, including announcers, clients and indexers, do possess the necessary keys to decrypt a given message, the message should decrypted and validated as if it were a non-encrypted message.
-
-### Non-Encrypted Messages
-
-Validating correctness for a non-encrypted message will consist of verifying that each field in the message is defined and meets the formatting rules listed in the subsections below for the given field type.
+Validating correctness for a message will consist of verifying that each field in the message is defined and meets the formatting rules listed in the subsections below for the given field type.
 
 #### Addresses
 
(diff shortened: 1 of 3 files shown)


pages/Messages/Validation.md
```diff
@@ -0,0 +1,173 @@
+---
+name: Validation
+route: /Messages/Validation
+menu: Messages
+---
+
+# Message Validation
+
+All messages read from an archive should be validated to guarantee the message's authenticity, authorization and legality.
```
wilwade, line 9: > legality

That's an odd word to use here to me.

pages/Messages/Validation.md
```diff
@@ -0,0 +1,173 @@
...
+
+```
+"😀", "🤌🏼", "👩🏻‍🎤", "🧑🏿‍🏫", "🏳️‍🌈", "🏳️‍⚧️", "⚛︎", "🃑", "♻︎"
+```
+
+Additionally, none of the following should be considered valid:
+
+```
+"F", ":custom-emoji:", "<custom-emoji>", "ᚱ", "ᘐ", "״"
+```
```
wilwade, line 74: I thought we wanted to allow for arbitrary length emoji to allow for merged emoji?

pages/Messages/Validation.md
```diff
@@ -0,0 +1,173 @@
...
+
+##### Type Enumerators
+
+1. Type Enumerator fields must be a valid enumerator value as defined in the [Message Overview](/Messages/Overview).
+
+##### URI Fields
+
+1. URI fields must include meet all standards defined in [RFC3986](http://www.ietf.org/rfc/rfc3986.txt).
+1. URI fields must not refer to localhost or any reserved IP addresses as defined in [RFC6890](https://datatracker.ietf.org/doc/html/rfc6890)
+1. URI fields must use either a `http` or `https` protocol. Support for other protocols may be added in the future.
```
wilwade, line 92: I think we want to ban http and only allow https for now.

pages/Messages/Validation.md
```diff
@@ -0,0 +1,173 @@
...
+1. URI fields must not exceed 1024 bytes.
+
+### Announcement Authenticity
+
+Validation of announcement authenticity will consist of verifying the message's signature using the appropriate public key for the user listed in the `fromAddress` of the message and the [Secp256k1](https://en.bitcoin.it/wiki/Secp256k1) algorithm, as defined in the [Message Signatures](/Messages/Signatures) specification.
```
wilwade, line 97: I think we should get more detailed here.

`fromAddress` I think is going to translate into fromId (being the DSNP Id). So the steps to are:

1. Resolve the fromId to a identity contract
2. using ecrecovery get the public key from the signature
3. Test the public key against the identity contract with IDelegation.isAuthorizedTo

pages/Messages/Validation.md
```diff
@@ -0,0 +1,173 @@
...
+## Type Specific Validation
+
+Specific types of DSNP messages will also include their own specific validations depending on the semantic intent of the message type.
+These specific validations follow:
+
+### Broadcasts
+
+1. All broadcast messages with a `"type"` field of `"Follow"` in the activity pub content must include a date in the `"published"` field preceded by a follow [Graph Change](/Messages/Overview#GraphChange) DSNP message with no unfollow events between the time of the follow and the broadcast messages.
+1. All broadcast messages with a `"type"` field of `"Like"` or `"Dislike"` in their activity pub content must be preceded by a [Reaction](/Messages/Overview#Reaction) DSNP message referring to the same object, i.e. that the `URI` message field of DSNP message identified by the `inReplyTo` field of the reaction is identical to the `"object"` field of the like or dislike activity pub content.
+1. All broadcast messages with a `"type"` field of `"Undo"` must include an `"object"` field with either a valid DSNP identifier referring to a previously announced broadcast message or a URI matching the URI of a previously announced broadcast event.
```
wilwade, line 128: I wonder if we should have native support for this as well...

pages/Messages/Validation.md
```diff
@@ -0,0 +1,173 @@
...
+name: Validation
+route: /Messages/Validation
+menu: Messages
+---
+
+# Message Validation
+
+All messages read from an archive should be validated to guarantee the message's authenticity, authorization and legality.
+Messages should be validated before being written to an archive file, however clients should not rely on this being the case and should perform their own validation on messages in the interest of protection against malicious actors.
+
```
shannonwells, line 11: It just occurred to me that there might be legal liability ramifications in making statements like this and the one above.  We might want to submit this to PL's legal department for review.  I think we're going to need a ToS soon with a "no warranty" exclusion sort of thing.

pages/Messages/Validation.md
shannonwells: I say ban it forever.

pages/Messages/Validation.md
```diff
@@ -0,0 +1,173 @@
...
+1. Content must be valid a valid JSON object as defined in [RFC7159](https://datatracker.ietf.org/doc/html/rfc7159).
+1. Content must include a type field matching one or more of the [core types](https://www.w3.org/TR/activitystreams-vocabulary/#h-types) or [extended types](https://www.w3.org/TR/activitystreams-vocabulary/#h-extendedtypes) defined in the [Activity Vocabulary](https://www.w3.org/TR/activitystreams-vocabulary/) specification.
+1. Content must include all necessary fields associated with the given type defined in the [Activity Pub](https://www.w3.org/TR/activitypub/) specification.
+
+Additional fields not required or defined by the Activity Pub specifications may also be included in accordance with various extensions of the specification, such as [Mastodon](https://docs.joinmastodon.org/spec/activitypub/), [ForgeFed](https://github.com/forgefed/forgefed) or one of the many [potential future extensions](https://www.w3.org/wiki/ActivityPub_extensions) proposed by the [W3C](https://www.w3.org).
+
+If the content of a message is no longer accessible, i.e. the URI of the message returns a 404 or 500 HTTP status, the message should no longer be considered valid.
```
shannonwells, line 110:
```suggestion
If the content of a message is no longer accessible, i.e. the URI of the message returns a 404 or 500 HTTP status, the message is invalid.
```

pages/Messages/Validation.md
```diff
@@ -0,0 +1,173 @@
...
+It is also recommended that implementations provide a warning either in the console or directly to the user with the associated HTTP status.
+For example, a message such as `"Content Inaccessible: Error 404"` would suffice.
+
+### Content Authenticity
+
+Authentication of a message's contents should be verified by hashing the exact contents of the body returned by the message URI and comparing it with the `contentHash` field of the given message.
```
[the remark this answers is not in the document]
shannonwells: Or at least a reference to a document about how the hashing works.

pages/Messages/Validation.md
```diff
@@ -0,0 +1,173 @@
...
+
+## Type Specific Validation
+
+Specific types of DSNP messages will also include their own specific validations depending on the semantic intent of the message type.
+These specific validations follow:
+
+### Broadcasts
+
+1. All broadcast messages with a `"type"` field of `"Follow"` in the activity pub content must include a date in the `"published"` field preceded by a follow [Graph Change](/Messages/Overview#GraphChange) DSNP message with no unfollow events between the time of the follow and the broadcast messages.
+1. All broadcast messages with a `"type"` field of `"Like"` or `"Dislike"` in their activity pub content must be preceded by a [Reaction](/Messages/Overview#Reaction) DSNP message referring to the same object, i.e. that the `URI` message field of DSNP message identified by the `inReplyTo` field of the reaction is identical to the `"object"` field of the like or dislike activity pub content.
```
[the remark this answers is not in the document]
shannonwells: Agreed I'm not really sure why there is a specific call-out for Like/Dislike -- aren't we allowing any reaction that is describable/referencable (for lack of better words) using our Reaction format?

pages/Messages/Validation.md
```diff
@@ -0,0 +1,173 @@
...
+
+## Optional Validations
+
+In addition to the aforementioned general validations on all messages and type specific validation on particular messages, there are also a set of optional validations that will be defined below and implementers are encouraged to support, however users, archivists and indexers may choose to ignore depending on their legal jurisdiction and individual preference.
+As such, it is strongly encouraged to make these features configurable by the end user in all implementations as allowable by law.
+
+### Blocklists & Allowlists
+
+Users may maintain private lists of other users who they do not wish to publicly block but choose to ignore all messages from.
+Likewise, archivists and indexers may also choose to ignore messages from users with repeated patterns of malicious or other objectionable behavior.
```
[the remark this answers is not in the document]
shannonwells: This could also have legal ramifications. Maybe it could be considered as destroying evidence, but I don't know.  This also seems like something we should run by a legal department.  This line of thought only occurred to me as I started reading through this PR.

pages/Messages/Validation.md
rosalinekarr: Fixed

pages/Messages/Validation.md
```diff
@@ -0,0 +1,173 @@
...
+In general, all implementations should validate all announcements for correctness against the [DSNP message schema](/Messages/Overview) and authenticity using [their provided signatures](/Messages/Signatures).
+Additionally, all client implementations should validate content for correctness against the [Activity Pub schema](https://www.w3.org/TR/activitypub/) and authenticity using the provided content hash.
+However, archivist implementations may choose to skip content validation steps in the interest of performance.
+
+### Announcement Correctness
+
+Validation of announcement correctness will vary depending on which fields are present on a particular DSNP message type.
+In general, correctness validation will consist of confirming that all necessary fields are present and that values in each field use a format appropriate for the field type.
+
+#### Encrypted Messages
```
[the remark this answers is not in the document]
rosalinekarr: I was thinking it would be relevant to distinguish encrypted vs. non-encrypted messages here because we're under the "Announcement Correctness" subsection. The "Announcement Authenticity" subsection below this one goes into more detail regarding signatures.

pages/Messages/Validation.md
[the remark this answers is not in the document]
rosalinekarr: My thinking was that users may or may not want to publish their like and dislike activity. This was actually a controversial change on Twitter a few years ago when they started including other people's likes in user's feeds by default. It wasn't long afterward that users demanded the ability to toggle the feature off and on.

Since activity pub already includes "Like" and "Dislike" object types, I figured we could just treat a published like as a broadcast message following a reaction message.

pages/Messages/Validation.md
rosalinekarr: Yeah, it's probably worth talking over in the next spec discussion meeting.

pages/Messages/Validation.md
```diff
@@ -0,0 +1,173 @@
...
+If the content of a message is no longer accessible, i.e. the URI of the message returns a 404 or 500 HTTP status, the message should no longer be considered valid.
+It is also recommended that implementations provide a warning either in the console or directly to the user with the associated HTTP status.
+For example, a message such as `"Content Inaccessible: Error 404"` would suffice.
+
+### Content Authenticity
+
+Authentication of a message's contents should be verified by hashing the exact contents of the body returned by the message URI and comparing it with the `contentHash` field of the given message.
+Given that the signature of the message is valid, this should serve as proof that the signing user posted the activity pub content.
+
+## Type Specific Validation
```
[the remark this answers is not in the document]
rosalinekarr: Yeah, I tried to be clear in my usage of "content" and "message" to indicate which I was referring to. Is there any lines in particular which aren't clear?

pages/Messages/Validation.md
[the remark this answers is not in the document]
rosalinekarr: Absolutely, we should run this by legal. For what it's worth, I meant to say announcer here instead of archivist, but I supposed the same concerns apply.

pages/Messages/Validation.md
```diff
@@ -0,0 +1,173 @@
...
+
+1. All reply and reaction messages must include an `inReplyTo` field identifying an existing DSNP message.
+1. All reply and reaction messages must be from and signed by a user, the replier, for which the user who posted the original content referred to in the `inReplyTo` field has never previously published a broadcast message with a activity pub content `"type"` field of `"Block"` and `"object"` field containing the string `dsnp://` followed by the user id of the replier, thereby indicating that the original poster has previously blocked the replier.
+
+## Optional Validations
+
+In addition to the aforementioned general validations on all messages and type specific validation on particular messages, there are also a set of optional validations that will be defined below and implementers are encouraged to support, however users, archivists and indexers may choose to ignore depending on their legal jurisdiction and individual preference.
+As such, it is strongly encouraged to make these features configurable by the end user in all implementations as allowable by law.
+
+### Blocklists & Allowlists
```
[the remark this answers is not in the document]
rosalinekarr: I feel like we should have it somewhere in the spec since it's going to be a common concern, but this particular file might not be the best place.

pages/Messages/Validation.md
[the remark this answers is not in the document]
wilwade: Switching it to announcer clears all of my concerns. No one is required to announce content they disagree with (in line with no one is required to host your website), but you can do it yourself. Archivists have (potentially) a much more neutral role to play.

pages/Messages/Validation.md
wilwade: Hmm... I think that works. I do wonder if we want to move some of this into the announcement type spec. We might have some reorg to do once some of the dust settles.

pages/Messages/Validation.md
```diff
@@ -0,0 +1,117 @@
...
+* All assumptions from [DSNP Signatures](/Messages/Signatures)
+
+## Announcement Correctness
+
+Validation of announcement correctness will vary depending on which fields are present on a particular DSNP message type.
+Validating correctness for a message will consist of verifying that each field in the message is defined and meets the formatting rules listed in the subsections below for the given field type.
+
+#### Addresses
+
+1. Address fields must be exactly 20 bytes.
```
wilwade, line 36: I actually don't know if we have any addresses anymore as we are changing `fromAddress` to be `dsnpId` (which is just 64 bits (8 bytes))

wilwade (dismissed):
A few questions, but looking nice and simple!

pages/Messages/Validation.md
```diff
@@ -0,0 +1,117 @@
...
+
+Additionally, none of the following should be considered valid:
+
+```
+"F", ":custom-emoji:", "<custom-emoji>", "ᚱ", "ᘐ", "״"
+```
+
+#### Hashes
+
+1. Hash fields must be exactly 32 bytes.
```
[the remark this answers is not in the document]
rosalinekarr: Since the hash is a byte value, I don't think we need to specify hexadecimal. As long as the column in the batch file takes raw bytes, users ought to be able to pass around hashes in whatever format they want, hex, base64, base53, etc.

pages/Messages/Validation.md
rosalinekarr: ok, I'll drop this section and add a "User Id" section.

pages/Messages/Validation.md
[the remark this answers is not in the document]
wilwade: Hmm... I guess it matters for the signature serialization, but you're right, it doesn't matter as long as parquet is storing bytes not strings. I glanced quickly and it would just be a byte array, so I think we are fine.

wilwade (approved):

PR #68, opened 2021-05-21
Bugs/update signing spec for eth rpm limitations 177926856

rosalinekarr:
Problem
=======
The standard Ethereum RPC provided by MetaMask and other wallets does not include a `signDigest` method, only a `sign` method which automatically prefixes all messages
[#177926856](https://www.pivotaltracker.com/story/show/177926856)

Solution
========
Update our signatures spec to include this prefix so we can use wallet signing

### DIFF
--- a/pages/Messages/Signatures.md
+++ b/pages/Messages/Signatures.md
@@ -34,7 +34,7 @@ The process for generating signatures for DSNP announcements consists of three d
 To serialize messages, objects must first be converted to a string then prepended with the standard [Ethereum RPC prefix](https://eth.wiki/json-rpc/API#eth_sign).
 To convert encrypted messages to a string, the bytes of the message should be encoded in [base 64](https://developer.mozilla.org/en-US/docs/Glossary/Base64).
 To convert non-encrypted messages to a string, each key of the DSNP object should be concatenated with its value in alphabetical order.
-Once converted, the string must be prefixed with "\x19Ethereum Signed Message:\n" and the length of the string.
+Once converted, the string must be prefixed with "\x19Ethereum Signed Message:\n" and the byte length of the string.
 
 Once the data is serialized, the serialized string should be hashed using [keccak-256](https://en.wikipedia.org/wiki/SHA-3).
 This fixed length hash generated can then be signed using [secp256k1](https://google.com/search?hl=en&q=secp256k1) and the publishing user's private key.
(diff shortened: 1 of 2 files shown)


pages/Messages/Signatures.md
```diff
@@ -51,28 +53,28 @@ For example, given the following DSNP broadcast message:
...
 ```
-contentHash0x67890fromAddress0x12345urihttps://www.projectliberty.io/
+\x19Ethereum Signed Message:\n\x69contentHash0x67890fromAddress0x12345urihttps://www.projectliberty.io/
 ```
 
 This would be the expected hash:
 
 ```
-0x45c42591e0154088325055f26664002bf05f211db9e5d7c7bfc588dc309698e0
+0x4c233fc1d71f92e4b8b7408291d45b5aa2ec3689d14091abc9207c4eca3886ad
```
wilwade, line 62: I get `0x6ad0d59e6d5f7aaee5998e5d584d8c55a3c01f09bf0f5b4c49b2399eca22a82a` when running this through manually and through https://web3js.readthedocs.io/en/v1.3.4/web3-eth-accounts.html?highlight=recover#hashmessage which should result in the same hash.

I think it might be that the length is supposed to be in decimal not hex.

pages/Messages/Signatures.md
```diff
@@ -51,28 +53,28 @@ For example, given the following DSNP broadcast message:
...
 
 Verifying announcements can be done by repeating the serialization and hashing steps from the signing process then validating the generated hash against the publishing user's public key.
 The user's public key can be fetched from the Identity contract as described in the [Identity spec](/Identity/Overview).
 
-Given the message and signature provided in the previous example, the following public key should return true when verifying:
+Given the message and signature provided in the previous example, the following public key should return a matching hash when verifying:
 
 ```
-+IUGiZsGHq+3s/Zgxl8TQMRgydvsOX1hUwMzHykoqGw=
+0x033dd49eb9790f8473399093c10eead398bbd76eae0905b9b26fd550aced63e858
```
wilwade, line 79: This should be a public key right?

wilwade (changes requested):

shannonwells (commented):
Looks good to me aside from already requested changes.  I have nothing to add.

pages/Messages/Signatures.md
rosalinekarr: This is a public key. I just encoded it as hex instead of base64 for consistency.

pages/Messages/Signatures.md
rosalinekarr: Fixed

pages/Messages/Signatures.md
```diff
@@ -51,28 +53,28 @@ For example, given the following DSNP broadcast message:
...
+0xf9d3010a7b1f199b5ea58473908c0c78a5cb04fdead2a05a7961c9a7ff87f1e81c654605b50f65b7d255177b0b49bbd97fa07fb88f9a7ea90565baa96b68e61e
 ```
 
 ## Verifying Messages
 
 Verifying announcements can be done by repeating the serialization and hashing steps from the signing process then validating the generated hash against the publishing user's public key.
 The user's public key can be fetched from the Identity contract as described in the [Identity spec](/Identity/Overview).
 
-Given the message and signature provided in the previous example, the following public key should return true when verifying:
+Given the message and signature provided in the previous example, the following public key should return a matching hash when verifying:
```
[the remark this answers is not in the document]
rosalinekarr: Fixed

pages/Messages/Signatures.md
```diff
@@ -31,8 +31,10 @@ The process for generating signatures for DSNP announcements consists of three d
 1. Hash the serialized string
 1. Sign the hash
 
-For encrypted messages, the encrypted bytes of the `dsnpData` field can be hashed directly without any serialization.
-For non-encrypted messages, each key-value pair in the `dsnpData` object should be concatenated in alphabetical order with no separator characters.
+To serialize messages, objects must first be converted to a string then prepended with the standard [Ethereum RPC prefix](https://eth.wiki/json-rpc/API#eth_sign).
+To convert encrypted messages to a string, the bytes of the message should be encoded in [base 64](https://developer.mozilla.org/en-US/docs/Glossary/Base64).
```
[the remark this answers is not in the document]
rosalinekarr: Fixed

pages/Messages/Signatures.md
rosalinekarr: Fixed

wilwade (approved):
🚢

Issue #73, opened 2021-05-21
Future supported message types


aneyzberg:
### Future Supported DSNP message types
There are several DSNP message types that are not yet currently supported but will be supported at some point in the future

#### Drop

A dead drop message.

| dsnpData field | description | type | bloom |
| ------------- |------------- | ---- | --- |
| deadDropID | The Dead Drop ID (See [DeadDrops](TBD) | bytes32 | YES
| uri  | content uri  |  string  | no
| contentHash | keccak-256 hash of content |  bytes32 | no

#### KeyList, PrivateGraphKeyList, EncryptionKeyList

A KeyList rotation.

| dsnpData field | description | type | bloom |
| ------------- |------------- | ---- | --- |
| fromId | DSNP ID | bytes8 | YES
| keyList | new list of valid keys | bytes[] | no

#### EncryptedInbox

An encrypted direct message.
This describes the format once decrypted.
Possibly combine both of these and expect that all Inbox messages are encrypted.

| dsnpData field | description | type | bloom |
| ------------- |------------- | ---- | --- |
| toAddress | ID of the recipient | bytes20 | YES
| fromId | DSNP ID  | bytes8 | YES
| contentHash | keccak-256 hash of content | bytes32 | no
| uri  | content uri  | string | no
Tell us how to reproduce this issue.

aneyzberg:
There are other messaging formats we considered before deciding on the format in [Messaging Overview](/Messages/Overview). T

There potential formats are described below.

### Variable Announcement Format
* All actions are posted to chain with some or all pertinent information about the action
* Different information is posted depending on the action.
* **Advantages**
    * can index all actions without requesting off-chain data
    * can validate most actions without requesting off-chain data
    * can archive some actions without requesting off-chain data
* **Disadvantages**
    * more data (likely more costly up front) than a simple URI

#### Some other options:

* Emit no topic, have a single contract that subscribers watch for events from.  Subscribers can perform filtering based on the `dsnpTopic` field.
* The topic is the dsnpType (possibly not an enum).  Subscribers could listen for desired topics.
### Unified Announcement Format

This is currently not the recommended solution, but is presented as a comparison.

* All actions are posted to the chain with some pertinent information about the action
* The same information is posted regardless of action
* The rest of the information for the action is stored off chain.
* **Advantages**
    * less data on chain --> lower up-front costs and lower node storage requirements
    // right to be forgotten is unclear
[...]

PR #76, opened 2021-05-21
clean up messages overview page

aneyzberg:
Problem
=======
Improve inconsistencies in overview page of Message spec to chip away at weekly spec chore
[link to Pivotal Tracker #178195764](https://www.pivotaltracker.com/story/show/178195764)

Solution
========
Clean up overview page to make it much more clear.

Change summary:
---------------
* Rename `fromAddress` to `fromId` in spec
* Remove potential message type implementations and create an issue in Github - https://github.com/LibertyDSNP/spec/issues/73
* Remove types that will not be supported in the future. 
* Remove options that were considered but not implemented and add as comment on github issue


pages/Messages/Overview.md
```diff
@@ -175,83 +99,22 @@ A visual reply to a post.
...
-
-### No data except hashes on chain
-
-Only hashes of the events are stored on chain; everything else is interpreted via API(s)
+#### Other notes
+1. Only some Liberty Platform activities will have the full message posted to chain. Examples:
+    * Public messages (profile changes, public posts, reactions)
+    * Direct messages between accounts (dead drop, inbox)
+    * Key changes (private graph KeyList, encryption KeyList, public key rotation)
+    * Graph changes (follow/unfollow)
```
wilwade, line 120: I think we can drop this section. I think it is more confusing than leaving it out. Also the current list of examples appears to just be everything

pages/Messages/Overview.md
```diff
@@ -112,61 +78,19 @@ A public reply post.
 | ------------- |------------- | ---- | --- |
 | inReplyTo | ID of the announcement the reply references |  bytes32 | YES
```
wilwade, line 79: I think these inReplyTos are incorrect now. They should be string style identifiers: https://spec.projectliberty.io/Messages/Identifiers

pages/Messages/Overview.md
```diff
@@ -175,83 +99,22 @@ A visual reply to a post.
 | dsnpData field | description | type | bloom |
 | ------------- |------------- | ---- | --- |
 | inReplyTo | ID of the message the reaction references |  bytes32 | YES
```
wilwade, line 101: I think these inReplyTos are incorrect now. They should be string style identifiers: https://spec.projectliberty.io/Messages/Identifiers

wilwade (changes requested):
Two small notes

wilwade (approved):
✨ clean

PR #81, opened 2021-06-02
Add createAndRegisterBeaconFactory function

acruikshank:
Problem
=======
Currently, new users must make two sequential transactions to create an identity: the first to create a proxy and the second to register a handle using the address of the proxy. Sequential transactions require more gas, more time, and more complicated error handling code on the client.

Solution
========
This PR introduces a `createAndRegisterBeaconProxy` method on the beacon factory that creates the proxy and registers it in one step. This function simply calls `createBeaconProxyWithOwner` and then `Registry.register`.

Change summary:
---------------
* Add `createAndRegisterBeaconProxy` signature.


wilwade (approved):
🚢

PR #82, opened 2021-06-03
Remove the indexed flag on handles for DSNPRegistryUpdate

wilwade:
Problem
=======
The spec and the interface were out of sync. An indexed string in an event is unable to recover the original string. So DSNPRegistryUpdate was not able to fill need to retrieve the original handle.
[link to Pivotal Tracker #178266672](https://www.pivotaltracker.com/story/show/178266672)

Options:
1. Don't index the handle, meaning that you cannot search for historical owners of handles without pulling the entire log history.
2. Rename indexed handle to searchHandle or handleHash and add a new non-indexed field of handle

Solution
========
Option 1: Remove the indexed flag on handles for DSNPRegistryUpdate

Change summary:
---------------
* Remove index from handle in DSNPRegistryUpdate Event
* Replace resolveHandleToAddress and resolveHandleToId with just resolveRegistration

### DIFF
--- a/pages/Identity/Factory.md
+++ b/pages/Identity/Factory.md
@@ -221,7 +221,7 @@ interface IIdentityBeaconFactory {
     function createBeaconProxyWithOwner(address beacon, address owner) external returns (address);
 
     /**
-     * @dev Creates a new identity with the adddress as the owner and registers it with a handle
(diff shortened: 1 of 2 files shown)


aneyzberg (approved):
lgtm

Issue #58, opened 2021-05-17
Activity Streams Suggested Support Levels


wilwade:
We should spec a minimum of support for Activity Streams Vocabulary and related extensions

Activity Streams Vocab is so large, I don't think any project has support for all of it.

Levels of support or feature support?

- Minimum
- Text
- Images
- Video
- Other?

Should result in:
- A part of the spec that sets the minimum expected level of support for Activity Streams Vocab (potentially split out by feature)
- Discussion of questions around Activity Streams support

Activity Streams Vocabulary: https://www.w3.org/TR/activitystreams-vocabulary/

PR #89, opened 2021-06-16
Features/message ordering 178248599

rosalinekarr:
Problem
=======
We don't have any clear standards on ordering messages for display in a feed.
[#178248599](https://www.pivotaltracker.com/story/show/178248599)

Solution
========
Added an activity pub spec with a requirement for timestamps on all activity pub objects.

I also added a section for ordering DSNP messages since it was called out in the story, but I expect it will be significantly changed in [#178435322](https://www.pivotaltracker.com/story/show/178435322).


pages/Messages/Ordering.md
```diff
@@ -0,0 +1,29 @@
...
+1. Describe a standard for ordering DSNP messages and resolving conflicts in DSNP message with time dependent context.
+
+## Assumptions
+
+* All assumptions from [DSNP Messages](/Messages/Overview)
+
+## Ordering
+
+DSNP messages must be ordered at the highest level by the block number of the block in which a message's batch is announced with higher block numbers indicating more recent messages and lower block numbers representing older ones.
+After block number, announcements published in the same block must be ordered by the transaction index of the block with higher transaction indices indicating more recent messages and lower transaction indices indicating older ones.
```
acruikshank, line 26: Should this be "transaction index of the transaction", "transaction index of the announcement", or just "transaction index"?  "transaction index of the block" doesn't make any sense to me.

pages/Messages/Ordering.md
```diff
@@ -0,0 +1,29 @@
...
+
+1. Describe a standard for ordering DSNP messages and resolving conflicts in DSNP message with time dependent context.
+
+## Assumptions
+
+* All assumptions from [DSNP Messages](/Messages/Overview)
+
+## Ordering
+
+DSNP messages must be ordered at the highest level by the block number of the block in which a message's batch is announced with higher block numbers indicating more recent messages and lower block numbers representing older ones.
```
acruikshank, line 25: Throughout this section it would be good to substitute "older" with "sorted first" ("sorted earlier"?) and newer with "sorted last" to emphasize this isn't a strictly temporal search.

acruikshank (approved):
Some minor concerns, but this seems technically correct and well written.

pages/Messages/Ordering.md
```diff
@@ -0,0 +1,29 @@
...
+After block number, announcements published in the same block must be ordered by the transaction index of the block with higher transaction indices indicating more recent messages and lower transaction indices indicating older ones.
+After transaction index, announcements within the same transaction must be ordered by log index with higher log indices representing more recent messages and lower log indices indicating older ones.
+After log index, batch files announced in the same announcement must be ordered in the reverse order of how they appear in the `announce` solidity method call with batch files appearing first being the oldest and last being the most recent.
```
wilwade, line 28: We forced batch files to be one per log event

rosalinekarr: [`Announce()` takes an array.](https://github.com/LibertyDSNP/contracts/blob/main/contracts/Announcer.sol#L10) Doesn't that mean multiple files can be announced in the same log event or is there something I'm missing?

pages/Messages/Ordering.md
rosalinekarr: I didn't want to say first or last since clients may choose to sort content by newest or oldest. It would make sense for a client to display a broadcast feed in reverse chronological order and display a conversation in forward chronological order.

pages/Messages/Ordering.md
rosalinekarr: Fixed

pages/ActivityPub/Overview.md
```diff
@@ -0,0 +1,55 @@
...
+name: Overview
+route: /ActivityPub/Overview
+menu: ActivityPub
+---
+
+# Activity Pub Overview
+
+Content shared via the DSNP will generally consist of URIs pointing to documents containing Activity Pub JSON objects.
+For the purposes of the DSNP, only a subset of the [Activity Pub Streams](https://www.w3.org/TR/activitypub/) and [Activity Pub Vocabulary](https://www.w3.org/TR/activitystreams-vocabulary/) specifications must be supported by default.
+Optionally, some implementers may choose to support more of the Activity Pub standards as they see fit, but as a warning, these standards may include ambiguities which could lead to different and incompatible implementations of certain features.
```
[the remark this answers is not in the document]
rosalinekarr: Fixed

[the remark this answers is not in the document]
rosalinekarr: Also, I think there should be a longer conversation about potential incompatibilities and further restrictions we could place on the Activity Pub standard, but I figured it was out of scope for this particular story. This task was just getting the timestamp requirement in to unblock UsApp and example client development.

rosalinekarr:
@wilwade: Feel free to merge on my behalf while I'm out on vacation.

pages/Messages/Ordering.md
wilwade: It can be announced in one transaction (covered by the line above), but not one log event.

https://github.com/LibertyDSNP/contracts/blob/main/contracts/IAnnounce.sol#L11

pages/ActivityStreams/Overview.md
```diff
@@ -0,0 +1,55 @@
...
+| 0.1     | Tentative |
+
+## Purpose
+
+1. Identify what aspects of the existing Activity Streams specification is supported, encouraged or enforced in the DSNP.
+1. Specify any extensions or modifications of the Activity Streams specification that must be supported or required by the DSNP.
+
+## Assumptions
+
+* All broadcast messages link to activity stream JSON, not any other form of content.
```
acruikshank, line 30: I don't understand why this is an assumption. Why can't we assert it in the spec? If it's merely an assumption it's likely to be wrong. The header for this section could be "Links".

pages/ActivityStreams/Overview.md
```diff
@@ -0,0 +1,55 @@
...
+Messages not meeting these requirements must be ignored.
+
+#### Timestamps
+
+All Activity Streams content must include a `"published"` key with a timestamp value string matching the [ISO8601](https://www.iso.org/iso-8601-date-and-time-format.html) format with timezone.
+Clients may choose to treat content marked as published significantly in the future as scheduled content which will be presented to the user later or ignored entirely.
+Likewise, clients may choose to treat content marked as published significantly in the past as archival content which is only displayed to the user when sought out or ignore it entirely.
+
+In any case, timestamps cannot be trusted to represent the true time that a message is posted on the DSNP due to the decentralized nature of the system and the relativistic limits of communication across the world.
+Clients may sort content by it's provided published timestamp, but it should be made clear to the user where ever possible that these timestamps are no necessarily accurate.
```
acruikshank, line 55: "no necessarily accurate" should be "not necessarily accurate"

pages/Messages/Ordering.md
acruikshank: The notion of a sort order implies the notion of first and last, and "earlier" or "later" in the sort. It's unfortunate that a lot of our language to describe sorting revolves around time. The point of this comment is that, to the extent possible, it would be good to separate the notion of sort order from temporal order. So "sorted first" or "sorted earlier" is better that "older", since items that appear earlier in the sort aren't necessarily older in any meaningful sense. "Sorted higher" and "sorted lower" might be better still. How the sort order is used to display items should be irrelevant.

pages/Messages/Overview.md
```diff
@@ -38,8 +38,8 @@ Some announcements, such as broadcasts and replies, will contain references to c
 
 #### Content
 
-Content items, or simply content, refers to an activity pub compliant object hosted at some URI and intended to be posted via an announcement in a batch.
-Generally, content within the DSNP specification will be defined by the [W3C Activity Pub specification](https://www.w3.org/TR/activitypub/#object-without-create), however extensions, such as the [Mastodon Activity Pub specification](https://docs.joinmastodon.org/spec/activitypub/) may also be implemented.
+Content items, or simply content, refers to activity streams compliant JSON hosted at some URI and intended to be posted via an announcement in a batch.
```
acruikshank, line 41: Not related to this PR, but I'm wondering if we should be referring to URLs instead of URIs since there is an expectation that they represent locations where content can be retrieved.

acruikshank (approved):
Looking good. There's one grammatical mistake that needs to be corrected.

pages/Messages/Overview.md
wilwade: URI vs URL is a debate that continues to rage on :) Let's discuss for a future PR.

pages/ActivityStreams/Overview.md
wilwade: That's a good point. These are not assumptions, they are important reminders of requirements elsewhere in the spec. Thoughts on how to title it? Links doesn't quite capture that to me, but it is better than Assumptions.

shannonwells: This title seems just fine. I don't understand why the title makes it an assumption. It's called a "requirement" in the title.  Did this change after these comments were posted?  Did I miss something?  The only thing I'd change is  to put a "must" in both bullet points, i.e. "must link"

pages/ActivityStreams/Overview.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Linking Requirements
+
+* All broadcast messages link to activity stream JSON, not any other form of content.
+* All reply messages link to an activity stream JSON, not any other form of content.
+
+## Minimum Subset
+
+At minimum, all clients implementing the DSNP must support presentation for the following message types from the Activity Streams vocabulary: `"Link"`, `"Note"` and `"Person"`.
+It is also strongly encouraged to support `"Audio"`, `"Image"`, `"Mention"`, `"Profile"`, `"Tombstone"` and `"Video"` where ever possible.
```
shannonwells, line 36: For this we should perhaps explain why this is strongly encouraged.  What happens if it's not supported?

pages/ActivityStreams/Overview.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Extensions
+
+In addition to the standards defined by the Activity Streams specifications, some extensions of Activity Streams are also required by the DSNP due to its decentralized context.
+
+### Required Keys and Values
+
+Certain suggested properties of objects in the Activity Streams standard are required for validation by the DSNP.
```
shannonwells, line 44: It would be useful to link to where these required properties are specified in this spec.

shannonwells (approved):
ok well there's my 2 cents.  nothing blocking

pages/ActivityStreams/Overview.md
wilwade: I like this idea. Having these be listed out and linked here. @rosalinekarr perhaps you can catch it Monday.

pages/ActivityStreams/Overview.md
wilwade: Added

pages/ActivityStreams/Overview.md
wilwade: Actually this could be covered by https://github.com/LibertyDSNP/spec/issues/58

PR #90, opened 2021-06-23
Spec/more activity pub 178020655

rosalinekarr:
Problem
=======
The Activity Streams standard is largely open-ended which may lead to incompatibilities between various clients implementing the DSNP.
[#178020655](https://www.pivotaltracker.com/story/show/178020655)

Solution
========
Define minimum and recommended feature support for the Activity Streams standard in our spec.


wilwade (changes requested):

pages/ActivityStreams/Overview.md
```diff
@@ -33,22 +33,347 @@ These deviations from the standard will be thoroughly documented here.
...
+  "height": 2250
+}
+```
+
 ## Extensions
 
 In addition to the standards defined by the Activity Streams specifications, some extensions of Activity Streams are also required by the DSNP due to its decentralized context.
 
-### Required Keys and Values
+### Content Contexts
```
wilwade, line 354: I'm not sure what I think about this.

pages/ActivityStreams/Overview.md
```diff
@@ -33,22 +33,347 @@ These deviations from the standard will be thoroughly documented here.
...
+
+Clients may sometimes want to represent content as existing within a specific presentation or social context, such as Reddit's subreddit concept, Slack's channels or old-style forum rooms.
+It is recommended that implementers represent this context in activity objects with the `"context"` field.
+If present, this field must have a string value starting with the name of the implementing service followed by forward slash and a string unique to the specific context.
+Implementers are encouraged to add additional forward slash separators to their unique string to represent contexts within contexts.
+For example, a `#development` channel context within a `LibertyDSNP` organization on Slack could be represented by the string, `"Slack/LibertyDSNP/#development"`.
+
+Clients may choose to ignore activity objects with contexts outside the user's current context, but it is strongly recommended that all contexts be made accessible to the user through some form of navigation.
+
+### Deletion Support
```
wilwade, line 364: We should talk about this. It puts an additional stress on content checking. If the content doesn't match the hash, you would then have to check for deleted field data. Is that what we want? Also how do we validate that the deletion was user requested?

pages/ActivityStreams/Overview.md
```diff
@@ -33,22 +33,347 @@ These deviations from the standard will be thoroughly documented here.
...
+### Content Contexts
+
+Clients may sometimes want to represent content as existing within a specific presentation or social context, such as Reddit's subreddit concept, Slack's channels or old-style forum rooms.
+It is recommended that implementers represent this context in activity objects with the `"context"` field.
+If present, this field must have a string value starting with the name of the implementing service followed by forward slash and a string unique to the specific context.
+Implementers are encouraged to add additional forward slash separators to their unique string to represent contexts within contexts.
+For example, a `#development` channel context within a `LibertyDSNP` organization on Slack could be represented by the string, `"Slack/LibertyDSNP/#development"`.
+
+Clients may choose to ignore activity objects with contexts outside the user's current context, but it is strongly recommended that all contexts be made accessible to the user through some form of navigation.
+
```
wilwade, line 363: Close to contexts, we should likely go ahead an specify and support [tags](https://www.w3.org/TR/activitystreams-vocabulary/#dfn-tag) and therefore Mentions ala https://www.w3.org/TR/activitystreams-vocabulary/#microsyntaxes

wilwade (commented):
few other notes

pages/ActivityStreams/Overview.md
```diff
@@ -33,22 +33,347 @@ These deviations from the standard will be thoroughly documented here.
...
+Clients are also recommended to support `"video/H256"` complying with [RFC7798](https://www.iana.org/go/rfc7798), `"video/mp4"` complying with [RFC4337](https://www.iana.org/go/rfc4337) and `"video/raw"` complying with [RFC4175](https://www.iana.org/go/rfc4175).
+
+Video objects may also include a `"duration"` field with a string value complying with the [XML Schema 11-2](https://www.w3.org/TR/xmlschema11-2/) standard for duration strings as recommended in the [Activity Vocabulary](https://www.w3.org/TR/activitystreams-vocabulary/#dfn-duration) specification.
+Video objects linking to live streaming content may use the string `"Live"` for the duration field or omit it entirely.
+
+Video objects may include a `"preview"` field with an activity sub-object of type link, a URL string or an array of sub-objects and/or strings referring to shorter video files or image files provided as a preview of longer video files.
+It is recommended that implementers at minimum include one image previews wherever possible.
```
acruikshank, line 316:
```suggestion
Implementers should include at least one image preview.
```

pages/ActivityStreams/Overview.md
```diff
@@ -33,22 +33,347 @@ These deviations from the standard will be thoroughly documented here.
...
+Preview images should generally include at least one file of the following resolutions: `320x240`, `640x480` or `1024x768`.
+
+Clients that support video objects should also include fields of `"height"` and `"width"` each with positive integer value on either the root activity object or the individual link sub-objects if multiple are present.
```
acruikshank, line 319: Again, why is this optional and if it needs to be optional, why specify it?

pages/ActivityStreams/Overview.md
```diff
@@ -33,22 +33,347 @@ These deviations from the standard will be thoroughly documented here.
...
+If these fields are not provided, clients may attempt to infer the media type from the headers of the linked file or simply ignore the object.
+Clients may also ignore video files of exceptionally large or small sizes, however videos with a preview of one of the prior mentioned resolutions must always be supported.
```
acruikshank, line 321: The preview field might also be exceptionally large or small. I'm not sure we want to make this a requirement.

pages/ActivityStreams/Overview.md
```diff
@@ -33,22 +33,347 @@ These deviations from the standard will be thoroughly documented here.
...
+
 ## Extensions
 
 In addition to the standards defined by the Activity Streams specifications, some extensions of Activity Streams are also required by the DSNP due to its decentralized context.
 
-### Required Keys and Values
+### Content Contexts
+
+Clients may sometimes want to represent content as existing within a specific presentation or social context, such as Reddit's subreddit concept, Slack's channels or old-style forum rooms.
+It is recommended that implementers represent this context in activity objects with the `"context"` field.
```
acruikshank, line 358: Implementers should represent

pages/ActivityStreams/Overview.md
```diff
@@ -33,22 +33,347 @@ These deviations from the standard will be thoroughly documented here.
...
+Implementers are encouraged to add additional forward slash separators to their unique string to represent contexts within contexts.
+For example, a `#development` channel context within a `LibertyDSNP` organization on Slack could be represented by the string, `"Slack/LibertyDSNP/#development"`.
+
+Clients may choose to ignore activity objects with contexts outside the user's current context, but it is strongly recommended that all contexts be made accessible to the user through some form of navigation.
+
+### Deletion Support
+
+Users may occasionally want to delete an activity object by changing it's contents, thereby changing it's hash value and invalidating any signature associated with it via broadcast or reply messages on chain.
+Likewise, hosts of activity objects may want to do the same for the purpose of removing illegal or otherwise objectionable content.
+Implementers are strongly recommended to represent this by adding a `"deleted"` field to the hosted activity object with an [ISO8601](https://www.iso.org/iso-8601-date-and-time-format.html) timestamp representing the time at which the content was removed.
```
acruikshank, line 369: What does strongly recommended mean? Can we just say "must"? Why would implementers not be able to do this?

pages/ActivityStreams/Overview.md
```diff
@@ -33,22 +33,347 @@ These deviations from the standard will be thoroughly documented here.
...
+Clients are also recommended to support `"image/gif"` complying with [RFC2045](https://www.iana.org/go/rfc2045) and `"image/heic"` complying with the [ISO/IEC JTC-1](http://www.iso.org/iso/jtc1_home.html) standard where possible.
+
+Image objects may include a `"preview"` field with an activity sub-object of type link, a URL string or array of sub-objects and/or strings referring to a smaller resolution image or images provided as previews of larger image files.
+Preview fields should generally include at least one file of the following resolutions: `320x240`, `640x480` or `1024x768`.
+
+Clients that support image objects should also include fields of `"height"` and `"width"` each with positive integer value either on the root activity object or the individual link sub-objects if multiple are present.
+If these fields are not provided, clients may attempt to infer the media type from the headers of the linked file or simply ignore the object.
+Clients may also ignore image files of exceptionally large or small sizes, however images with a preview of one of the prior mentioned resolutions must always be supported.
```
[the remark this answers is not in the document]
rosalinekarr: My thinking was that this should be optional since height and width aren't always clearly defined in the case of unusual image formats like SVG or panoramic MP4.

acruikshank (approved):
There's a lot of really good content here. I've added many suggestions on mostly stylistic topics, but I have some more general concerns:

1. This part of the spec dips a bit into the presentation of content (what must or must not be presentable, etc.). This seems a little out of scope for a protocol spec, and forces some rather arbitrary decisions. The spec could be (should be?) limited to what fields the activity objects must contain and which ones they can contain, and what it means for these fields to be valid.
2. The language is generally pretty loose. There's not much justification for why some things are required and some are merely recommended. Some general framework how implementers should make decisions about required, recommended and optional content would be welcome.

pages/ActivityStreams/Overview.md
rosalinekarr: We don't. My thinking was that this is just a nice to have for clients, if they want to support it, but none of it is required or verified. If users want to completely delete an activity pub file without providing a reason, they can, and if they want to provide a timestamp and reason, they can too.

This is just an optional feature so clients can display messages like "Content Deleted on Jan 1, 1970 by Author" and avoid confusion in replies and such.

pages/ActivityStreams/Overview.md
rosalinekarr: We also [support tags](https://github.com/LibertyDSNP/spec/pull/90/files#diff-9213ad9daeed82825a98eec5ea7f3c9e7a48c946b6b372b46e5d1e3f50d24cb7R93) but this is a different use case. Tags are many-to-many, which is great for drawing other user's attention to stuff, but it's not great for chat-room-style conversations like Slack or Discord channels. Contexts are stricter in that a post can only ever have one or none.

pages/ActivityStreams/Overview.md
```diff
@@ -33,22 +33,347 @@ These deviations from the standard will be thoroughly documented here.
...
+It is also strongly encouraged to support `"Audio"`, `"Image"`, `"Profile"` and `"Video"` where ever possible.
 Clients may ignore messages that they do not support.
 
+### Required Activity Types
+
+#### Link
+
+Link type objects must be supported to represent linked content posted by users.
+Link objects must have a `"type"` field with a value of `"Link"` and a `"href"` field with a valid URL string value.
+Clients must at least support valid URLs with a `"https://"` protocol, but clients may choose to ignore any other protocols as they see appropriate for the security of users.
```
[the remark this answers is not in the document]
rosalinekarr: See previous discussion: https://github.com/LibertyDSNP/spec/pull/90#discussion_r658013090

pages/ActivityStreams/Overview.md
```diff
@@ -33,22 +33,347 @@ These deviations from the standard will be thoroughly documented here.
...
+}
+```
+
+#### Note
+
+Note type objects must be supported to represent text content posted by users.
+Note objects must have a `"type"` field with a value of `"Note"` and a `"content"` field with a string value.
+Clients should support display of content with unicode text as possible in their given browser or operating system.
+Clients may also support display of [Markdown](https://daringfireball.net/projects/markdown/) or [BBCode](https://www.bbcode.org) as appropriate.
+If a note contains Markdown, BBCode or some other formatting, it is recommended that an `"mediaType"` field be added to the object with a [MIME type](https://www.iana.org/assignments/media-types/media-types.xhtml) string value representing the type of encoding present, i.e. `"text/markdown"`.
```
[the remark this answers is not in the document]
rosalinekarr: I think my reasoning for this being a recommendation instead of a requirement is because it's not really enforceable as a requirement. There's no way I could distinguish markdown text in a note's content that doesn't have a `mediaType` field from plain text that doesn't need a `mediaType` field, so I wouldn't be able to mark those messages as invalid and ignore them.

I suppose we could say it's a requirement here, but there's no validation function we could write to back it up.

pages/ActivityStreams/Overview.md
```diff
@@ -33,22 +33,347 @@ These deviations from the standard will be thoroughly documented here.
...
+Clients are also recommended to support `"image/gif"` complying with [RFC2045](https://www.iana.org/go/rfc2045) and `"image/heic"` complying with the [ISO/IEC JTC-1](http://www.iso.org/iso/jtc1_home.html) standard where possible.
+
+Image objects may include a `"preview"` field with an activity sub-object of type link, a URL string or array of sub-objects and/or strings referring to a smaller resolution image or images provided as previews of larger image files.
+Preview fields should generally include at least one file of the following resolutions: `320x240`, `640x480` or `1024x768`.
+
+Clients that support image objects should also include fields of `"height"` and `"width"` each with positive integer value either on the root activity object or the individual link sub-objects if multiple are present.
```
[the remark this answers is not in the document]
rosalinekarr: See prior discussion: https://github.com/LibertyDSNP/spec/pull/90#discussion_r658089018

pages/ActivityStreams/Overview.md
```diff
@@ -33,22 +33,347 @@ These deviations from the standard will be thoroughly documented here.
...
+
+```json
+{
+  "@context": "https://www.w3.org/ns/activitystreams",
+  "type": "Audio",
+  "name": "Super Great Podcast",
+  "url": {
+    "type": "Link",
+    "href": "https://example.org/podcast.mp3",
+    "mediaType": "audio/mp3"
```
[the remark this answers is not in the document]
rosalinekarr: **Notes following meeting:**

- We will add hashes for files.

shannonwells (commented):
The one thing I want to add -- my overall thinking about this, and it's just occurred to me, is that if I'm an implementer reading a spec, and I want to know what's required and what's just recommended/optional, I would love a much easier to scan bullet point format, or a table summary, rather than having to do a text search for a bunch of different synonyms like (recommended|should|optional) or (required|must|will|...)

pages/ActivityStreams/Overview.md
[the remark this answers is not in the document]
rosalinekarr: **Notes following meeting:**

- Height & Width must be included
- Remove preview recommendation entirely

pages/ActivityStreams/Overview.md
```diff
@@ -33,22 +33,347 @@ These deviations from the standard will be thoroughly documented here.
...
+  "duration": "PT3H12M"
+}
+```
+
+#### Image
+
+Clients are recommended to support presentation for activity objects of type image to represent image content posted by users.
+Image type objects, if supported, must at minimum contain a `"type"` field with the value `"Image"` and a `"URL"` with either a URL string or an activity sub-object of type link containing a URL pointing to an image file.
+
+Image objects are also recommended to include a `"mediaType"` field on their link sub-object with a standard [MIME type](https://www.iana.org/assignments/media-types/media-types.xhtml) string value.
```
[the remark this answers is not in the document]
rosalinekarr: Required

pages/ActivityStreams/Overview.md
rosalinekarr: Preview recommendations removed.

pages/ActivityStreams/Overview.md
rosalinekarr: My thinking was that we can't require this because there's no way to prevent users from invalidating their activity pubs without adding a deleted timestamp.

I'll change it to "should."

pages/ActivityStreams/Overview.md
```diff
@@ -79,13 +81,15 @@ And implementers may choose to ignore the following link objects:
...
-Note objects must have a `"type"` field with a value of `"Note"` and a `"content"` field with a string.
-Clients should support content with unicode text as possible in their given browser or operating system.
+Note objects must include a `"type"` field with a value of `"Note"`.
+Note objects must include a `"content"` field with a string.
+
+Clients may support content with unicode text as possible in their given browser or operating system.
 Clients may support [Markdown](https://daringfireball.net/projects/markdown/), [BBCode](https://www.bbcode.org) or other formatting standards.
-If a note contains Markdown, BBCode or some other formatting, it is recommended that a `"mediaType"` field be added to the object with a [MIME type](https://www.iana.org/assignments/media-types/media-types.xhtml) string representing the type of encoding present, i.e. `"text/markdown"`.
+If a note contains Markdown, BBCode or some other formatting, the note object should include a `"mediaType"` field with a [MIME type](https://www.iana.org/assignments/media-types/media-types.xhtml) string representing the type of encoding present, i.e. `"text/markdown"`.
```
jvasile, line 89: Under what conditions would a note that isn't text/plain *not* need a mimetype specified?  Maybe this should be required?

Do we want to limit notes to text?  "If specified, the mediaType field must contain a media type with a top level of 'text'"

pages/ActivityStreams/Overview.md
```diff
@@ -32,23 +32,354 @@ These deviations from the standard will be thoroughly documented here.
...
+  "duration": "PT10M32S",
+  "mediaType": "video/webm"
+}
+```
+
 ## Extensions
 
-In addition to the standards defined by the Activity Streams specifications, some extensions of Activity Streams are also required by the DSNP due to its decentralized context.
+In addition to the standards defined by the Activity Streams specifications, some extensions of Activity Streams are required by the DSNP due to its decentralized context.
+
```
shannonwells, line 353: This part is confusing to me, because it says some extensions are required for the protocol, yet the fields listed seem to actually be optional.  So is the support for them required, but the fields themselves are optional?

shannonwells (approved):
Nothin but nits

pages/ActivityStreams/Overview.md
rosalinekarr: Reworded

pages/ActivityStreams/Overview.md
rosalinekarr: Changed to must

pages/ActivityStreams/Overview.md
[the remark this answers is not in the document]
rosalinekarr: On second thought, I've updated this to be a requirement.

pages/ActivityStreams/Overview.md
```diff
@@ -6,7 +6,7 @@ menu: Activity Streams
 
 # Activity Streams Overview
 
-Content shared via the DSNP will generally consist of URIs pointing to documents containing Activity Streams JSON objects.
+Content shared via the DSNP will generally consist of URLs pointing to documents containing Activity Streams JSON objects.
```
wilwade, line 9: We need to get standard on URI vs URL or what we mean by each...

pages/ActivityStreams/Overview.md
```diff
@@ -32,23 +32,354 @@ These deviations from the standard will be thoroughly documented here.
...
+    "href": "https://upload.wikimedia.org/wikipedia/commons/c/c0/Big_Buck_Bunny_4K.webm",
+    "width": 4000,
+    "height": 2250
+  },
+  "duration": "PT10M32S",
+  "mediaType": "video/webm"
+}
+```
+
+## Modifications and Extensions
```
wilwade, line 350: We don't discuss the absence of the "id" field. Should we?

pages/ActivityStreams/Overview.md
rosalinekarr: I changed this at @acruikshank's request: https://github.com/LibertyDSNP/spec/pull/90#discussion_r657971374

should I change it back?

pages/ActivityStreams/Overview.md
```diff
@@ -32,23 +32,354 @@ These deviations from the standard will be thoroughly documented here.
 
 ## Minimum Subset
 
-At minimum, all clients implementing the DSNP must support presentation for the following message types from the Activity Streams vocabulary: `"Link"`, `"Note"` and `"Person"`.
-It is also strongly encouraged to support `"Audio"`, `"Image"`, `"Mention"`, `"Profile"`, `"Tombstone"` and `"Video"` where ever possible.
+All clients implementing the DSNP must support presentation for the following message types from the Activity Streams vocabulary: `"Link"`, `"Note"` and `"Person"`.
```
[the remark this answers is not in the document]
rosalinekarr: We do that with the non-required types like `Audio` and `Video`. My thinking with requiring these three is that they're necessary for the use of the DSNP. Even a photo-only app would need `Person` to represent users and `Link` for the photos. I suppose we could drop `Note`, but I can't really imagine any apps that wouldn't want text replies.

pages/ActivityStreams/Overview.md
rosalinekarr: Added a section

pages/ActivityStreams/Overview.md
```diff
@@ -32,23 +32,354 @@ These deviations from the standard will be thoroughly documented here.
...
+Clients may choose to represent deleted content in presentation with a small tombstone message, such as "content removed," or ignore deleted content entirely.
+Implementers wishing to specify a tombstone message for deleted content may change the `"type"` of the object to `"Tombstone"` and add or update the `"content"` field to include a message explaining the deletion, i.e. "removed due to DMCA violation" or "removed by author."
+
+### Tags
+
+Users may choose to include tags in activity objects to bring them to the attention of other users.
+Implementers must support a `"tag"` field on any activity object which may contain a string, an activity sub-object of type `"Tag"` or `"Mention"` or an array of strings and/or sub-objects.
+Mention sub-objects must include a `"type"` field with the value `"Mention"` and a `"href"` field containing the DSNP User Id of the mentioned user.
+
+### Timestamps
```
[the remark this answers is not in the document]
rosalinekarr: I added the `"published"` requirement to each section.

I'm not really sure about rearranging the documents. Let's discuss it more in the next spec meeting.

pages/ActivityStreams/Overview.md
```diff
@@ -32,23 +32,354 @@ These deviations from the standard will be thoroughly documented here.
...
+  "duration": "PT3H12M"
+}
+```
+
+#### Image
+
+Clients should support image activity objects to represent image content posted by users.
+Image objects must include a `"type"` field with the value `"Image"`.
+Image objects must include a `"URL"` with either a URL string or an activity sub-object of type `"Link"` containing a URL pointing to an image file.
+Image objects must include a `"hash"` field with a [keccak-256](https://en.wikipedia.org/wiki/SHA-3) hash of the body of the linked image file.
```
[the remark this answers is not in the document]
rosalinekarr: 1. Added.
2. Added a "Content Proofs" section to the extensions and linked it as appropriate.

wilwade (approved):
Approved. Moved reorg to new issue: https://github.com/LibertyDSNP/spec/issues/96

pages/ActivityStreams/Overview.md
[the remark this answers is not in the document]
harry-evans: I know I am late to this, but could we make mediaType a "may", but say if not present, it must be interpreted as "text/plain" (or some other default type)? That gives a well defined path for handling things if mediaType is absent, vs the message just being invalid.

Issue #96, opened 2021-07-01
Discussion: Spec Reorg


wilwade:
From Weekly Spec Discussion 2021-07-01

We are wanting to reorganize the spec to try to make the different pieces more independent, but also more clear which are a part of "Core" DSNP and which are "Extensions" to DSNP.

Extension: You don't have to do any of this, but if you do, here's the spec on how to do it.

For example. Not everyone needs to support video content. But there is a spec on how to do video content if you do it.

This issue is to propose and discuss ideas on how to do the reorg.

wilwade:
Ok. Here is a general suggestion for how things could work.

In the weekly spec discussion (2021-07-08) we had a naming discussion related to this:
- Batch -> Batch Publication
- Message / Announcement -> Announcement
- Content / Message Content -> Activity Content

So one now publishes a batch full of announcements that reference activity content.

So high level:
- Networks -> Cleanup
- Implementation Status -> Move to the index?
- Identity -> Cleanup, maybe remove factory info
- Graph -> Move to sub Announcements
- Messages -> Announcements
- Batches -> Batch Publications?
- Activity Streams -> Activity Content
- Archivists -> ?
- Reference Glossary -> Cleanup terms

## More Detailed Outline with Notes

- Networks
- Identity
- Batch Publications
  - Overview
  - Publishing (Previously Messages->Overview->Announcing DSNP Events)
  - Validation
  - File Format
    - Bloom filters
    - Range queries?
  - Queries (Remove/move to blog?)

- Announcements
  - Overview
  - Announcement Types (separate pages? That way each could have notes on the inverse/updates actions?)
    - Broadcast
      - has a contentUrl and a contentHash
      - Creates a DSNP Announcement Identifier
      - contentUrl must resolve to any Activity Content
    - Reply
      - has a contentUrl and a contentHash
      - Creates a DSNP Announcement Identifier
      - References DSNP Announcement Identifier
      - contentUrl must resolve to any Activity Content
    - Profile
      - has a contentUrl and a contentHash
      - Technically creates a DSNP Announcement Identifier, but do we want to support replies/reactions to it?
      - contentUrl must resolve to an Activity Content of type Profile
    - Graph
      - References Identity Spec
      - Move graph stuff under here
    - Reaction
      - References DSNP Announcement Identifier
  - Ordering
  - Validation
  - DSNP Announcement Identifier
  - Duplicate Handling (New section)
    - Replay Attacks

- Activity Content
[...]

PR #101, opened 2021-07-14
Revise the Batches section of the Spec to be Batch Publications

wilwade:
Problem
=======
The spec needs clearing and consistent language

Part 1 of #96

Solution
========
Cleanup and renaming of the Batch section of the spec

Change summary:
---------------
* Batches -> Batch Publications
* IAnnounce -> IPublish (and associated updates)
* Split up the Batch Overview
* Messages/Announce -> BatchPublications/Publish


pages/BatchPublications/FileFormat.md
```diff
@@ -10,45 +10,40 @@ Batch files are stored and transferred in Apache Parquet format.
 
 ## Specification Status
 
-| Version | Status |
----------- | ---------
-| 0.1     | Draft |
+| Version | Status   |
+| ------  | -------- |
+| 1.0     | Proposed |
```
aneyzberg, line 15: is 1.0 proposed or is it some other more permanent status? Also just a question - how are we determining status. Do we have that documented somehere?

aneyzberg (approved):
just a single question that is non-blocking. LGTM

pages/BatchPublications/FileFormat.md
wilwade: We don't have any official numbering yet, but I think it is reasonable to consider 1.0 for Betanet. We should discuss this with the team however.

Status names are here: https://spec.dsnp.org/ under "Specification status definitions"

PR #103, opened 2021-07-21
Updated all instances of hexadecimal in spec to use 0x prefix

rosalinekarr:
Problem
=======
We're currently inconsistent in our usage of prefixes on hexadecimal content because some of our underlying libraries use a `0x` prefix and some use no prefix at all.
[#178965517](https://www.pivotaltracker.com/story/show/178965517)

Solution
========
Use a 0x prefix everywhere and convert where non-prefixed hex is needed.

Change summary:
---------------
* Updated all hexadecimal requirements to include a `0x` prefix requirement.


wilwade (approved):
0xapprove

enddynayn (approved):

PR #106, opened 2021-07-22
Tombstone

### LINKED ISSUE #80: How do I delete content (and let people know)?
Deleting content is not addressed in the spec, but I think it should be.

Currently:
- Some things cannot be "deleted": Reactions
- Some things can be undone: Follow -> Unfollow
- Some things can be updated: Profile
- Some things cannot be deleted, but the referenced content can be: Broadcast, Reply (Although nothing asks caching services to remove it)

## Questions
- Should reactions be able to be undone?
- How can Broadcasts/Replies be retracted?
- How can edits propagate?

## Previous Discussion Thoughts
- Announcement type of "Metadata" that could include updated contentHash or tombstone flag
- 404 body content (although that wouldn't fix it for caches)
- Leverage some form of "self" moderation flag

### LINKED ISSUE #85: Reaction Redactions
# Summary

There should be some way users can remove a reaction after making one.
This issue lists a variety of approaches we could take to implementing this feature.

# Specification

## Option 1: New DSNPType

We have previously discussed adding a `Metadata` DSNP Message Type. If we did so, we could have specific metadata messages for indicating that a previously posted reaction should be ignored or removed.

## Option 2: Removed Flag

Alternatively, we could add a flag to the existing reaction DSNP Message Type indicating that a particular reaction is actually a reversal of a previous reaction. In this case, to remove a reaction, a user would post an identical reaction to the batch but with the removal flag set to true.

## Option 3: Anti-Emoji

Like anti-matter annihilating matter, we could extend the emoji unicode specification with the concept of anti-emoji. Anti-emoji would consist of a single `-` character followed by any valid emoji. Posting an anti-emoji would then indicate a removal of any previously posted emoji reaction consisting of the same string without the `-` character posted by the same user.

For example, if I mistakenly reacted `👍` to a post of someone sharing bad news, I could redact said reaction with another react message with the emoji string `-👍`. This would be notably distinct from `👎` which may carry a different semantic meaning depending on context.

# Copyright

Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/).

wilwade:
Problem
=======
We need to be able to undo things.

Closes #85
Closes #80 

Solution
========
Created a new announcement type that allows a user to tombstone

with most of the team in weekly spec meeting

Change summary:
---------------
* Added Tombstone Announcement Type

### DIFF
--- a/pages/Announcements/Overview.md
+++ b/pages/Announcements/Overview.md
@@ -14,14 +14,14 @@ All Announcements have a [signature](/Announcements/Signatures) to validate the 
 
 Each Announcement has a enumerated type for use when separating out a stream of Announcements.
 
-| Value | Name | Description | DSNP Announcement URI |
-|------ | ---- | ----------- | -------------------- |
-| 0 | Reserved | reserved for future use | - |
-| 1 | [Graph Change](/Announcements/Types/GraphChange) | social graph changes | no |
-| 2 | [Broadcast](/Announcements/Types/Broadcast) | a public post | YES |
-| 3 | [Reply](/Announcements/Types/Reply) | a public response to a Broadcast | YES |
-| 4 | [Reaction](/Announcements/Types/Reaction) | a public visual reply to a Broadcast | no |
-| 5 | [Profile](/Announcements/Types/Profile) | a profile | YES |
+| Value | Name | Description | DSNP Announcement URI | Tombstone Allowed |
+|------ | ---- | ----------- | --------------------- | ----------------- |
+| 0 | [Tombstone](/Announcements/Types/Tombstone) | an invalidation of another announcement | no | no |
+| 1 | [Graph Change](/Announcements/Types/GraphChange) | social graph changes | no | no |
+| 2 | [Broadcast](/Announcements/Types/Broadcast) | a public post | YES | YES |
+| 3 | [Reply](/Announcements/Types/Reply) | a public response to a Broadcast | YES | YES |
+| 4 | [Reaction](/Announcements/Types/Reaction) | a public visual reply to a Broadcast | no | YES |
+| 5 | [Profile](/Announcements/Types/Profile) | a profile | YES | YES |
 
 ## Value Serialization
 
@@ -50,12 +50,15 @@ Announcements in [Batch Publication Files](/BatchPublications/Overview) have an 
 The `DSNPBatchPublication` Ethereum events are ordered by information provided in the transaction.
 Announcements in a Batch Publication File are then ordered by row index.
 
-
 1. `DSNPBatchPublication` Block number ascending
 2. `DSNPBatchPublication` Transaction index ascending
 3. `DSNPBatchPublication` Log index ascending
 4. Batch Publication File Announcement row appearance order
 
+## Reverting an Announcement
+
+Announcements may not be deleted, but some may be marked as invalid by using a [Tombstone Announcement](/Announcements/Types/Tombstone).
+For example, if a user creates a reaction announcement, they may remove that reaction by creating a tombstone announcement.
 
 ## Non-Normative
 
--- a/pages/Announcements/Overview.md
+++ b/pages/Announcements/Overview.md
@@ -21,7 +21,7 @@ Each Announcement has a enumerated type for use when separating out a stream of 
 | 2 | [Broadcast](/Announcements/Types/Broadcast) | a public post | YES | YES |
 | 3 | [Reply](/Announcements/Types/Reply) | a public response to a Broadcast | YES | YES |
 | 4 | [Reaction](/Announcements/Types/Reaction) | a public visual reply to a Broadcast | no | YES |
-| 5 | [Profile](/Announcements/Types/Profile) | a profile | YES | YES |
+| 5 | [Profile](/Announcements/Types/Profile) | a profile | YES | no |
 
 ## Value Serialization
 
--- a/pages/Announcements/Types/Tombstone.md
+++ b/pages/Announcements/Types/Tombstone.md
@@ -56,9 +56,9 @@ It is NOT possible to revert a tombstone.
 
 ### targetSignature
 
-- MUST be an [Announcement Signature](/Announcements/Signatures) that the `fromId` has announced.
+- MUST be an [Announcement Signature](/Announcements/Signatures) that the `fromId` has announced
 
 ### signature
 
-- MUST be an [Announcement Signature](/Announcements/Signatures) over the all fields except this signature field.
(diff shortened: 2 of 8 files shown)


pages/Announcements/Overview.md
```diff
@@ -20,14 +20,14 @@ All Announcements have a [signature](/Announcements/Signatures) to validate the
...
-| 4 | [Reaction](/Announcements/Types/Reaction) | a public visual reply to a Broadcast | no |
-| 5 | [Profile](/Announcements/Types/Profile) | a profile | YES |
+| Value | Name | Description | DSNP Announcement URI | Tombstone Allowed |
+|------ | ---- | ----------- | --------------------- | ----------------- |
+| 0 | [Tombstone](/Announcements/Types/Tombstone) | an invalidation of another announcement | no | no |
+| 1 | [Graph Change](/Announcements/Types/GraphChange) | social graph changes | no | no |
+| 2 | [Broadcast](/Announcements/Types/Broadcast) | a public post | YES | YES |
+| 3 | [Reply](/Announcements/Types/Reply) | a public response to a Broadcast | YES | YES |
+| 4 | [Reaction](/Announcements/Types/Reaction) | a public visual reply to a Broadcast | no | YES |
+| 5 | [Profile](/Announcements/Types/Profile) | a profile | YES | YES |
```
rosalinekarr, line 30: I wonder if users should tombstone `Profile` announcements or just post new `Profile` announcements with their updated information?

pages/Announcements/Types/Tombstone.md
```diff
@@ -0,0 +1,65 @@
+---
+name: "Type: Tombstone"
+route: /Announcements/Types/Tombstone
+menu: Announcements
+---
+
+# Tombstone Announcement
+
+A Tombstone Announcement is a way to note that a previous announcement signature is invalid
+and the related Announcement should be considered reverted.
```
rosalinekarr, line 10: One sentence per line

pages/Announcements/Types/Tombstone.md
```diff
@@ -0,0 +1,65 @@
...
+| 4 | [Reaction](/Announcement/Types/Reaction |
+| 5 | [Profile](/Announcement/Types/Profile |
+
+### targetSignature
+
+- MUST be an [Announcement Signature](/Announcements/Signatures) that the `fromId` has announced.
+
+### signature
+
+- MUST be an [Announcement Signature](/Announcements/Signatures) over the all fields except this signature field.
```
rosalinekarr, line 65:
```
- MUST be the signature of the fromId user
```

rosalinekarr (approved):
LGTM, one minor fix and one non-blocking thought

pages/Announcements/Overview.md
```diff
@@ -56,12 +56,15 @@ Announcements in [Batch Publication Files](/BatchPublications/Overview) have an
...
-
 1. `DSNPBatchPublication` Block number ascending
 2. `DSNPBatchPublication` Transaction index ascending
 3. `DSNPBatchPublication` Log index ascending
 4. Batch Publication File Announcement row appearance order
 
+## Reverting an Announcement
+
+Announcements may not be deleted, but some may be marked as invalid by using a [Tombstone Announcement](/Announcements/Types/Tombstone).
+For example, if a user creates a reaction announcement, they may remove that reaction by creating a tombstone announcement.
```
aneyzberg, line 67: There was some ongoing discussion about this at some point. Did we decide that we would use Tombstones for Reactions?

aneyzberg (approved):
lgtm barring the open question about reaction types

enddynayn (approved):

pages/Announcements/Overview.md
shannonwells: I'm strongly in favor of extremely limited application of tombstone, i.e. it's only when a signature was invalid.  For example, I don't think an update to profile information should be done as a tombstone, if that's what you mean.  If you mean instead of allowing Tombstone on Profiles, and instead having people simply update their profile, I can see the usefulness of Tombstone as a way of indicating there was a problem. It might also help with forensics / diagnostics later (but maybe that's a weak argument).

pages/Announcements/Overview.md
shannonwells: Same here, I'd prefer that Tombstone is a revert of an invalid announcement to begin with, where invalid means a bad signature or the like. I think clients should have to handle things like accidental follow/unfollow clicks, by throttling, setting timeouts, use of confirm dialogs, etc.  Tombstoning should have a lot of friction.

shannonwells (approved):
Are we going to specify whether DSNP-compliant clients should display Tombstoned announcements?  Some services display the history of a post.  I expect that some clients will opt to display everything and mark content as Tombstoned (or maybe not if they're dishonest).

pages/Announcements/Types/Tombstone.md
wilwade: I thought the rule was max one sentence per line, thus splitting up a sentence was ok for readability. Thoughts?

pages/Announcements/Overview.md
wilwade: A "invalid announcement to begin with" would not need a tombstone as it is already invalid. A tombstone invalidates a previously valid signature.

That said I agree that clients ought to do all of the mentioned things like throttling. (However we should not note that in the specification. That would be a good implementation walk through as well as good to put into the Example Client)

pages/Announcements/Overview.md
wilwade: I'm for removing Profiles from the Tombstone list.

wilwade:
> Are we going to specify whether DSNP-compliant clients should display Tombstoned announcements? Some services display the history of a post. I expect that some clients will opt to display everything and mark content as Tombstoned (or maybe not if they're dishonest).

I don't think we need to specify this. (Although we should talk about what the example client should do)

pages/Announcements/Overview.md
shannonwells: > A "invalid announcement to begin with" would not need a tombstone as it is already invalid.
Right, of course. I should have said, "a retraction for something posted in error or, e.g. by a malicious person who  no longer has permissions on that account", not technically invalid.  It still should have a lot of friction.

wilwade: Ah that makes more sense.

The malicious person version should be done via delegation removal not via tombstoning however. The point of a tombstone is to say that something you did, you want to undo. I don't know of a way to increase the friction any more than just an announcement. (outside of PoW type stuff that is)

PR #114, opened 2021-08-19
Update serializations of numbers vs byte data

wilwade:
Problem
=======
#113 created new problems.

- While the serialization of the string for signing was clear, it was confusing what the core data types were.
- The parquet data types also caused some confusion as hexadecimal data would easily be encoded as a string byte array instead of properly as bytes

Solution
========
- Added a generic Data Type field for announcements
- Added a decimal numeric serialization for integers
- DSNP User Id is now decimal instead of hexadecimal
- hexadecimal no longer drops leading zeros as it represents just bytes instead of bytes and numbers
- Switched url and uris to use the UTF8 logical types for parquet
- Switched unsigned 64 bit integers to use UINT_64 for parquet and decimal for serialization
- Switched enums to use decimal for serialization

Note the UTF8 logical type is specified here: https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#string
Note the UINT_64 logical type is specified here: https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#deprecated-integer-convertedtype

with @rosalinekarr 

Also:
- Updated the second tier dependencies


shannonwells:
Let's be sure to put in a story to update the SDK for this, especially the validators if any changes need ot be made there

pages/Announcements/Overview.md
```diff
@@ -30,18 +30,36 @@ Most serializations use outside standards, but some require additional clarifica
...
 
+Used to represent bytes.
+
 - MUST use 0-9,a-f representation
 - MUST be lowercase
 - MUST be prefixed with a `0x`
-- MUST NOT have spaces
-- MUST NOT have any zero padding
+- MUST NOT have spaces or separators
+- MUST have two characters per byte
```
rosalinekarr, line 39: I wonder if we should say "MUST have two characters per byte in addition to the 0x characters"

pages/Announcements/Overview.md
```diff
@@ -30,18 +30,36 @@ Most serializations use outside standards, but some require additional clarifica
...
+| 32 | `0x3e34c4325f4461b9355027b314f3eb56d31af549f7da7bd9ef1ce951651e` | `0x00003e34c4325f4461b9355027b314f3eb56d31af549f7da7bd9ef1ce951651e` |
+
+### decimal
+
+Used to represent integers.
+Strings are used to avoid issues with different implementations of numbers.
+
+- MUST use 0-9 representation
+- MUST NOT have spaces or separators
+- MUST be a string
```
rosalinekarr, line 56: It might also be worth mentioning that the standard BigInt format with a trailing `n` is incorrect.

rosalinekarr (approved):
Only non-blocking feedback

pages/Announcements/Overview.md
rosalinekarr: I think we should at least add an example showing it's wrong.

pages/Announcements/Overview.md
shannonwells: No harm in being completely pedantic in one's language in a spec

pages/Announcements/Overview.md
wilwade: image

pages/Identifiers.md
```diff
@@ -7,13 +7,13 @@ route: /Identifiers
...
-- Is 8 bytes in size
+- 64 bit Unsigned Integer
 - MUST be registered in the [Identity Registry](/Identity/Registry)
-- MUST be serialized as [hexadecimal](/Announcements/Overview#hexadecimal)
+- MUST be serialized as [decimal](/Announcements/Overview#decimal)
 
 ## DSNP Content Hash
 
-- Is 32 bytes in size
+- MUST BE 32 bytes in size
```
acruikshank, line 16: nit: Capitalization is inconsistent here.

acruikshank (approved):
Looks great. Really like that we're treating the abstract data type at the same leve as the serializations.

pages/Announcements/Signatures.md
```diff
@@ -30,24 +30,24 @@ All parties interacting with Announcements should independently validate signatu
 
 ```json
 {
-  "announcementType": 1,
-  "fromId": "0x12345",
-  "contentHash": "0x67890",
+  "announcementType": "1",
```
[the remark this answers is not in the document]
enddynayn: I see your point. but i kinda like it.

enddynayn (approved):

pages/Announcements/Signatures.md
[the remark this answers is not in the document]
wilwade: I could do it as a table instead. That would be clearer without implying anything.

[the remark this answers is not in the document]
wilwade: Oh that is much better.

image

PR #119, opened 2021-09-20
Remove extra menu item

wilwade:
Problem
=======

Extra menu item was missed from when archivists were moved to drafts

Solution
========
Remove the extra menu item

Before:
image

After:
image


shannonwells (approved):
👍

Issue #127, opened 2021-11-15
Examples Typo in Activity Content: Associated Attachements


wilwade:
### Steps to reproduce
- https://github.com/LibertyDSNP/spec/blob/main/pages/ActivityContent/Associated/Attachments.md
- `"hash":` should always be an array according to the [hash spec](https://spec.dsnp.org/ActivityContent/Associated/Hash)

- [x] https://github.com/LibertyDSNP/spec/blob/main/pages/ActivityContent/Associated/Attachments.md?plain=1#L52
- [x] https://github.com/LibertyDSNP/spec/blob/main/pages/ActivityContent/Associated/Attachments.md?plain=1#L117
- [x] https://github.com/LibertyDSNP/spec/blob/main/pages/ActivityContent/Associated/Attachments.md?plain=1#L182

PR #128, opened 2021-11-16
Correct example mistake on Attachments

wilwade:
Problem
=======
Hash should always be an array of values. Issue #127

Solution
========
Correct the examples to show that they are arrays.

Change summary:
---------------
* Update three examples on the Attachments page

Steps to Verify:
----------------
1. "hash" must always be shown to be an array


shannonwells (approved):
Yes it should.

enddynayn (approved):

PR #134, opened 2022-01-26
design document document first pass

shannonwells:
Problem
=======
We want to create Design Documents for publication in our repo.
We need to create a process for generating Design Documents.
We need to describe how to write the Design Documents.

Solution
========
Write a Design Document meta-document.

with @raebot 

Change summary:
---------------
* First pass

### DIFF
--- a/DesignDocs.md
+++ b/DesignDocs.md
@@ -1,0 +1,57 @@
+# Design Documents
+
+## Purposes of this document
+* To describe what our design documents are for, how they are structured, and what content is in them.
+* To describe how and where DSNP Design Documents are produced and published.
+
+## What Is a Design Document?
+It's a document clearly articulating a problem, a solution to that problem, and why we think the proposal is the right solution.
+It has enough information to begin writing stories for implementing the solution in the document.
+ 
+## Audience
+The audience for DSNP Design Documents is mainly DSNP developers, both internal and at large.
+We also expect other developers and technically savvy readers will read these documents. 
+It's encouraged to link to technical explainers, such as Wikipedia, the DSNP Spec, or other DSNP Design Documents.
+This will help readers to learn about concepts in the proposal if need be.
+
+## Sections
+* Context + Scope: A short description of the landscape in which the new system is being built and what is actually being built
+* Problem Statement: The "why." A short summary of the issue(s) that this design solves.
+    This doesn't have to be a technical problem. 
+    "Developer unhappiness" or "improved user experience" are also problems.
+* Goals + Non-goals:  what this solution is trying to do, and is also _not_ trying to do, in concrete terms.
+* Proposal: A high level overview, followed by a detailed description.
+    This can include diagrams such as [A system context diagram](https://en.wikipedia.org/wiki/System_context_diagram)
+* Benefits and Risks: the reasons why this solution was chosen, and the risks this solution poses.
+    For example, the solution may be very simple, but there could performance bottlenecks above a certain threshold.
+    Another: the solution is well known and widely used, but it's not a perfect fit and requires complicated changes in one area.   
+* Alternatives + Rationale:  discuss alternatives that were considered, and why they were rejected. 
+    Note when there are absolute requirements that the solution does not and can't meet.
+    One example might be, it's a proprietary solution but we need something open source.
+* Sources: sources of information that led to this design.
+* Glossary (optional): if you think inline links to concepts are too distracting, include a glossary section. 
+    This can be links, text or both.
+
+## Process
+The process for adding a new DSNP Design Document is much like any other Pull Request in a code base.
+1. Create a branch in this repo.
+2. Commit your design doc in your branch, based on the guidelines here.
+3. Post the design document as a Draft PR for discussion.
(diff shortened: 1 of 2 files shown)


DesignDocs.md
```diff
@@ -0,0 +1,58 @@
...
+* Alternatives + Rationale:  discuss alternatives that were considered, and why they were rejected. 
+    Note when there are absolute requirements that the solution does not and can't meet.
+    One example might be, it's a proprietary solution but we need something open source.
+* Sources: sources of information that led to this design.
+* Glossary (optional): if you think inline links to concepts are too distracting, include a glossary section. 
+    This can be links, text or both.
+
+## Process
+The process for adding a new DSNP Design Document is much like any other Pull Request in a code base.
+1. Create a branch in this repo.
```
shannonwells, line 37: "This repo" is going to be a separate DesignDocs repo.

enddynayn (approved):

raebot (approved):
Looks great!

DesignDocs.md
```diff
@@ -0,0 +1,65 @@
...
+    Another: the solution is well known and widely used, but it's not a perfect fit and requires complicated changes in one area.   
+* **Alternatives and Rationale**: discuss alternatives that were considered, and why they were rejected. 
+    Note when there are absolute requirements that the solution does not and can't meet.
+    One example might be, it's a proprietary solution but we need something open source.
+* **Sources**: sources of information that led to this design.
+* **Glossary** (optional): if you think inline links to concepts are too distracting, include a glossary section. 
+    This can be links, text or both.
+
+## Process
+The process for adding a new DSNP Design Document is much like any other Pull Request in a code base.
```
raebot, line 39: Would we ever want to publish a draft design doc ahead of launching the related feature? Doing so could serve as an invitation for input from the public, which may or may not be desirable depending on the situation.

DesignDocs.md
```diff
@@ -0,0 +1,65 @@
...
+1. Create a branch in this repo.
+2. Commit your design doc in your branch, based on the guidelines here.
+3. Post the design document as a Draft PR for discussion.
+4. When the PR is accepted, it can be merged.
```
raebot, line 43: What is the timing between merging a design doc PR and launching its related feature? Would we merge the design doc PR ahead of launching the feature or would we want to wait? Would we encourage feedback from at-large contributors on the design doc when it's in the Draft PR state?

waylandli: I think that's a good point in terms of the design doc mirroring the implementation as close as possible. I'm just thinking about situations where a doc PR is approved but some changes were needed on the actual feature.

shannonwells: We'd post the Design Doc, it would be discussed, reviewed and merged, and then work would begin on it, starting with issues (stories) in the respective GitHub Project.  This would not be posted after the fact.

It doesn't preclude updates to the Design Docs, either.  I think it would be a case by case basis, but if for example it turns out that you don't need a certain API call or the call needs an extra parameter, it's updated in the Design Doc just the way a regular PR would be done in a code repo.

If say, the entire Design is scrapped because the feature is killed or something like that, we could maybe retire it by marking it retired, or something like that.  I think it's a little too early to lay out a process for that kind of thing.  We want to keep the first iteration simple and easy to pick up for now.

DesignDocs.md
```diff
@@ -0,0 +1,65 @@
+# Design Documents
+
+## Purposes of this document
+* To describe what our design documents are for, their structure, and what goes in them.
+* To describe how DSNP Design Documents are produced and where they're published.
+
+## What Is a Design Document?
```
wilwade, line 7: Do we need to talk about how this differs from a DIP? The goal is different in that a DIP is a specific change to the DSNP spec whereas this is more about ANY open source code we have. I could also see a design doc "producing" a DIP? Or perhaps this is just better than the DIP process for now?

sstruble: Good question. I'm ignorant of what the team has already outlined about a DIP and related process - please share anything if you can - but @jvasile @kfogel and I are meeting Monday to talk through a visual guide for governance and test cases. I don't know the right answer now other than keep it simple. I'll give a better suggestion on this soon.

shannonwells: The way I see it is a Design Doc is more directly related to writing code and not for the DSNP Improvement Proposals, although there's certainly overlap.

sstruble:
@shannonwells 
@kfogel and I created a new ["meta" repository](https://github.com/LibertyDSNP/meta) for guidelines that should apply across the entire project. I'm just starting to populate it, but when the team thinks this Design Doc is in good enough shape, I suggest that meta repo as a good landing place.

shannonwells:
Okay so we put this above into there, and then Design Docs relating to given repos go into the related repo.  That helps draw a better dividing line between Spec changes (such as DIPs) and Design Docs.

sstruble:
@shannonwells It might be easiest to just use GitHub's PR templates - revised, if required, for our needs - and point out in our contributing.md that we follow these automatically populate conventions. @kfogel  might have a thought here on the best approach

shannonwells:
Closing this and moving this to a PR against meta.

Issue #148, opened 2022-03-23
DIP-148 Rename DSNP Announcement URI to DSNP Content URI


wilwade:
# Abstract

While updating #145, we realized that the DSNP Announcement URI is a misnomer. It is instead a Content URI and should be identified as such in the spec.

# Motivation

- Clarity in names is important.

# Specification Pull Request

Current change pull request: TODO

## Additional Actions

- [ ] Create story in sdk-ts to update naming
- [ ] Create story in activity-content to update requirement
- [ ] Create story in activity-content-java to update requirement
- [ ] Create story in activity-content-swift to update requirement

# Rationale

The alternative of leaving it with a poor name made it hard to define.

# Backwards Compatibility

- Any sdk packages that use named parameters might need to be updated.

# Security Considerations

- Increases security in that it is a more precise name

# Dependencies

- #158

# References

- https://github.com/LibertyDSNP/spec/issues/145#issuecomment-1076659427

# Copyright

Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/).

shannonwells:
👍

Issue #145, opened 2022-03-07
DIP-145 Delegate Batch Signatures


wilwade:
# Abstract

A proposal to move the signature out of the individual announcement in a batch file and allow the implementation layer to choose how to validate batches.
Proposing for Ethereum, that the transaction with a batch publication should be signed by a delegate of the DSNP Ids in the batch.

## Terms

- *First Party*: The owner of the DSNP Id.
- *Second Party*: A delegate of the DSNP Id, but not the owner. Often a service provider.
- *Third Party*: An actor that is cryptographically disconnected from the DSNP Id(s).

# Motivation

- Simplicity in the spec
- Flexibility at the implementation layer
- (Ethereum) Increase in batch submission transparency

# Specification Pull Request

Current change pull request: https://github.com/LibertyDSNP/spec/pull/183

# Rationale

Half of the requirement in validation of an announcement was already at the implementation layer.
This proposal extends that to the entire question of how to validate an announcement.
This allows more freedom in the implementation to validate announcements in a variety of ways including at write-time instead of just read-time.

The other motivations draw from the implementation solution:

- Transparency increase. By cryptographically linking the batch to the delegate, users can know which delegates are honest and maintaining data availability.
- Batch file size is reduced.
- The work for validation is reduced to merely checking that the transaction sender is a delegate for each `fromId` in the batch.

# Backwards Compatibility

*This is a spec breaking change!*

- Announcement structure drops a field and validation is completely altered.
- Tombstones are broken! We would need to update Tombstones to use the `DSNP Announcement URI` likely.
- Therefore this DIP will require DSNP to increase the minor version.

### Migration Options:

1. Ignore previous data from DSNP.
2. Consider previous data from DSNP to be valid.
3. Have a migration block number per implementation.

For future changes like this, it would likely require 3.
For now, we can just do 1 or 2.

### Ethereum

No changes need to be made in Ethereum contracts to support this change.

### SDKs

SDKs will need to be upgraded to support the changes.

### Tombstone Announcement

Tombstone announcements will need to be updated to target something besides the signature

# Reference Implementation and/or Tests

- SDK Update/Branch: TBD.

# Security Considerations

This might appear to increase the level of trust that a user must have in a delegate, but the trust was just hidden before.
Previously, an app would need to either ask the user's trusted wallet to sign every announcement, have a key in the app, or have a backend service signing announcements.
The user's wallet option is secure, but not a great experience for the user.
[...]

PR #163, opened 2022-04-05
DIP-158 Require Published field on Activity Content type Note

### LINKED ISSUE #158: DIP-158 Make `published` required in Activity Content Type Note
# Abstract

While updating #145 and then #148 we realized that content hashes should be "required" to be unique, achievable via making the `published` field required.

# Motivation

- We want users to still be able to publish the same content more than once, but have replies, reactions, and other references be separated. (For example, "Good morning!")

# Specification Pull Request

Current change pull request: TODO

## Additional Actions

- [ ] Create story in activity-content to update requirement
- [ ] Create story in activity-content-java to update requirement
- [ ] Create story in activity-content-swift to update requirement

# Rationale

The other alternatives required more on chain data. This reduces the on chain data while still maintaining the same level of trust.

Requiring published on the `Profile` type does not appear to be necessary as it already uses a last profile wins resolution algorithm.

# Backwards Compatibility

- While this is a breaking change, as of this publishing no content exists without said value. It is already the effective default

# Security Considerations

While users could lie about the date, as duplicate Announcements are ignored, the purpose of generating unique hashes is as secure as it was before.

# Dependencies

None

# References

- https://github.com/LibertyDSNP/spec/issues/145#issuecomment-1076659427

# Copyright

Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/).

wilwade:
Problem
=======

While updating #145 and then #148 we realized that content hashes should be "required" to be unique, achievable via making the published field required.

Link to GitHub Issue(s): Closes #158

Solution
========
Marked `published` as required

Change summary:
---------------
- [x] Updated Spec Versions
- [x] Update `published`

Screenshots (optional):
-----------------------
image

### DIFF
--- a/.spelling
+++ b/.spelling
@@ -21,6 +21,7 @@ CouchDB
 DAO
 DIP
 DIP-148
+DIP-158
 DSNP
 DSNP.org
 DSNPArchives
@@ -158,6 +159,7 @@ namespace
 objectId
 parseable
 permissioned
+pre-1
 prepended
 prosumer
 punycode
--- a/pages/ActivityContent/Overview.md
+++ b/pages/ActivityContent/Overview.md
@@ -1,5 +1,5 @@
 # Activity Content Specification
-__Version 1.0.0__
+__Version pre-1.1.0__
 
 Content references shared via the DSNP consists of URLs pointing to documents containing Activity Streams JSON objects.
 For the purposes of the DSNP, restrictions are placed on the [Activity Streams 2.0](https://www.w3.org/TR/activitystreams-core/) specification.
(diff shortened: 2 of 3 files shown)


shannonwells (commented):
This is a breaking change, no?

wilwade:
> This is a breaking change, no?

@shannonwells It is for Activity Content, but not for any other specs.

shannonwells (approved):

saraswatpuneet (approved):

aramikm (approved):

Issue #157, opened 2022-03-28
Update CI Deployment for mdbook


mehtaishita:
## Acceptance Criteria
- [x] Update deploy workflow to use mdbook build
- [x] Should be dependent on CI lint passing
- [x] Should only deploy when tagged with a date based deployment regex `^\d{8}\.\d+`
- [x] Update readme to detail the date based deployment versioning.

## Notes
- Might be able to deploy from an artifact from the linting step

PR #164, opened 2022-04-05
enable tag based trigger

saraswatpuneet:
Problem
=======
## Acceptance Criteria
- [x] Update deploy workflow to use mdbook build
- [x] Should be dependent on CI lint passing
- [x] Should only deploy when tagged with a date based deployment regex `^\d{8}\.\d+`
- [x] Update readme to detail the date based deployment versioning.
Link to GitHub Issue(s): https://github.com/LibertyDSNP/spec/issues/157

Solution
========
- [x] ensure the cicd workflow for mdbook triggers on tag creation
- [x] step one should only trigger when date based tag is created via regex checks
- [x] deployment step should not trigger for PRs

Change summary:
---------------
- [x] Updated mdbook_ci.yaml to include three conditions above
- [x] Updated ReadMe with instructions to deploy via date tag

Steps to Verify:
----------------
1. PR checks should not run deployment step 
2.  Create  a temporary tag without following regex ^\d{8}\.\d+, build should not trigger in this case

Test:
- [x] Ran tests to ensure deployment workflow is picking up the correct sha of last commit on a branch

### DIFF
--- a/.github/workflows/mdbook_cd.yml
+++ b/.github/workflows/mdbook_cd.yml
@@ -23,7 +23,7 @@ jobs:
       - name: Decompress mdbook build artifact 📦
         run: mkdir book && tar -xvf spec_mdbook-${{ github.sha }}/spec_mdbook.tar ./book && ls ./book
 
-      # - name: Deploy mdBook to github pages 🚀
-      #   uses: peaceiris/actions-gh-pages@v3
-      #   with:
-      #     folder: ./book
+      - name: Deploy mdBook to github pages 🚀
+        uses: peaceiris/actions-gh-pages@v3
+        with:
+          folder: ./book
--- a/.github/workflows/mdbook_cd.yml
+++ b/.github/workflows/mdbook_cd.yml
@@ -21,7 +21,7 @@ jobs:
           check_artifacts:  true
 
       - name: Decompress mdbook build artifact 📦
-        run: mkdir book && tar -xvf spec_mdbook.tar ./book && ls ./book
+        run: mkdir book && tar -xvf spec_mdbook-${{github.sha}}/spec_mdbook.tar ./book && ls ./book
 
       # - name: Deploy mdBook to github pages 🚀
       #   uses: peaceiris/actions-gh-pages@v3
--- a/.github/workflows/mdbook_cd.yml
+++ b/.github/workflows/mdbook_cd.yml
@@ -21,8 +21,7 @@ jobs:
       - name: Decompress mdbook build artifact 📦
         run: tar -xvf spec_mdbook.tar ./book
 
-      # - name: Deploy mdBook to github pages 🚀
-      #   uses: JamesIves/github-pages-deploy-action@4.0.0
-      #   if: startsWith(github.ref, 'refs/tags/')
-      #   with:
-      #     folder: ./book
+      - name: Deploy mdBook to github pages 🚀
+        uses: peaceiris/actions-gh-pages@v3
+        with:
+          folder: ./book
--- a/.github/workflows/mdbook_cd.yml
+++ b/.github/workflows/mdbook_cd.yml
@@ -21,7 +21,7 @@ jobs:
           check_artifacts:  true
 
       - name: Decompress mdbook build artifact 📦
-        run: tar -xvf spec_mdbook.tar ./book | ls ./book
+        run: mkdir book && tar -xvf spec_mdbook.tar ./book && ls ./book
 
       # - name: Deploy mdBook to github pages 🚀
       #   uses: peaceiris/actions-gh-pages@v3
--- a/.github/workflows/mdbook_cd.yml
+++ b/.github/workflows/mdbook_cd.yml
@@ -3,7 +3,7 @@ name: Spec MDBook CD
 on:
   push:
     tags: 
-      - '*'
+      - '[0-9]+.[0-9]+'
 jobs:
   build:
     runs-on: ubuntu-20.04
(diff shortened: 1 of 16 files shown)


.github/workflows/mdbook_cicd.yml
```diff
@@ -1,7 +1,12 @@
...
-on: [pull_request]
-
+on:
+  push:
+    branches: [main]
+  create:
+    tags: 
+      - '^\d{8}\.\d+'
+  pull_request:
 jobs:
```
saraswatpuneet, line 10: there is a trigger when  a tag is created, tag should be dated and only tags that adhere to defined regex will be triggered

.github/workflows/mdbook_cicd.yml
```diff
@@ -43,9 +48,9 @@ jobs:
...
-      #   if: ${{ github.ref == 'refs/heads/main' }}
-      #   with:
-      #     github_token: ${{ secrets.GITHUB_TOKEN }}
-      #     publish_dir: ./book
+      - name: Deploy mdBook to github pages 🚀
+        uses: JamesIves/github-pages-deploy-action@4.0.0
+        # check tag
+        if: startsWith(github.ref, 'refs/tags/')
+        with:
+          folder: ./book
```
saraswatpuneet, line 56: on:
  create:
will ensure only tags following dated pattern trigger this workflow

here the if statement will ensure, deployment will only work for tags and not direct push to main or PR based trigger

.github/workflows/mdbook_cd.yml
```diff
@@ -0,0 +1,33 @@
...
+    runs-on: ubuntu-20.04
+
+    steps:
+      - name: Checkout 🛎️
+        uses: actions/checkout@v3
+
+      - name: Download mdbook build artifact 📦
+        uses: actions/download-artifact@v3
+        with:
+          name: spec_mdbook
```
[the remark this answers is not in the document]
saraswatpuneet: Updated CD to download the artifact tagged via sha and run deploy step

saraswatpuneet:
@wilwade   if it looks good then I can merge and test it out on main

saraswatpuneet:
also @wilwade  I do not think I have permissions to create tags on spec

.github/workflows/mdbook_cd.yml
```diff
@@ -0,0 +1,29 @@
...
+        uses: actions/checkout@v3
+
+      - name: Download mdbook build artifact 📦
+        uses: dawidd6/action-download-artifact@v2
+        with:
+          workflow: mdbook_ci.yml
+          workflow_conclusion: success
+          branch: main
+          check_artifacts:  true
+
```
saraswatpuneet, line 22: @wilwade  so here it will download artifacts on main branch so we get the latest commit on main

saraswatpuneet: This will also ensure tags creation always picks the main branch's latest artifacts @wilwade

.github/workflows/mdbook_cd.yml
```diff
@@ -0,0 +1,29 @@
+name: Spec MDBook CD
+
+on:
+  push:
+    tags: 
+      - '*'
+jobs:
```
saraswatpuneet, line 7: Lets trigger it for any tags creation since artifacts are picked from main branch latest commit

wilwade: We have other tags we use that we don't want to trigger for however. (like EVM-v1.1.0 or such) Those would also be targeting a main commit.

.github/workflows/mdbook_ci.yml
```diff
@@ -40,12 +44,6 @@ jobs:
       - name: Create mdbook build artifact 📦
         uses: actions/upload-artifact@v3
```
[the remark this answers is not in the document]
saraswatpuneet: Good point :+1:

.github/workflows/mdbook_cd.yml
saraswatpuneet: actions partially support some wild cards to filter (glob pattern but limited) ours regex will fall into [0-9]+.[0.9]+

saraswatpuneet: @wilwade  tested it it works so everytime a dated tag is created, deployment will trigger picking up last successful artifact zip from main

.github/workflows/mdbook_ci.yml
```diff
@@ -35,17 +39,13 @@ jobs:
         run: mdbook build
 
       - name: Create tarball 📦
+        if: ${{ github.ref == 'refs/heads/main' }}
         run: tar -cvf spec_mdbook.tar ./book
 
       - name: Create mdbook build artifact 📦
+        if: ${{ github.ref == 'refs/heads/main' }}
         uses: actions/upload-artifact@v3
```
saraswatpuneet, line 47: Only execute tar-balling and artifact upload steps for main branch

.github/workflows/mdbook_cd.yml
wilwade: @saraswatpuneet Does this mean we still release the latest version even if we tag a previous commit?

Imagining this line of commits from new to old:
E
C
B
A

If I tag C, does the build from E or C get used?

enddynayn (approved):

.github/workflows/mdbook_cd.yml
wilwade: @saraswatpuneet  just to double check. If I were to tag a commit that was one behind the head would it release the head or the tagged commit?

saraswatpuneet: @wilwade  good point as of now it only pics latest commit on main so it tag is behind head it will still go ahead and pick the head 

I am going to figure out how can I pick up the same commit hash as on tagged release

.github/workflows/mdbook_ci.yml
```diff
@@ -35,17 +39,13 @@ jobs:
...
         with:
-          name: spec_mdbook
+          name: spec_mdbook-${{ github.sha }}
           path: spec_mdbook.tar
```
saraswatpuneet, line 51: @wilwade  this will tag the artifact on main with its hash to during build time we can get the sha on commit for which tag was created and only download those artifacts and not the latest one

I think this should work but will know more once I merge to main

wilwade (approved):
👍

Issue #169, opened 2022-04-12
DIP-169 Block/Unblock Graph Changes


dopatraman:
# Abstract

A proposal to extend our protocol to include an announcement type for "blocking" and "unblocking" users.

# Motivation

- Accommodate the expectations of users from social networks that support this functionality

# Specification Pull Request

Current change pull request: #170 

# Rationale

Users of centralized social networks will use DSNP with certain expectations, one of which likely will be having the ability to block / unblock followers. One way to do this is extend our announcement type offering to include explicit Block and Unblock  types.

## Open Questions
- Is it possible to leverage ActivityStreams to accommodate this feature?
- Given that announcers have static IDs, will it be possible for users to block them? What implications does that have for how the entire network functions?
- Does this need to be a spec-level feature at all?

# Copyright

Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/).

# Reference 
https://docs.google.com/document/d/1ON2cJRvZYFWmTD6VzLXNnt2B4CijrdYfWap9JH6uw9E/edit

wilwade:
@dopatraman could you move the information from the Google Doc into the issue? Some of that should be discussed here.

Also I think we need to update the issue title to follow the DIP format.

wilwade:
Closing for now. We are going to think more about this with respect to private graph

PR #170, opened 2022-04-13
Added block/unblock graph change type for DIP-169

dopatraman:
Problem
=======
The purpose of this PR is to propose block/unblock functionality at a protocol level.

Link to GitHub Issue(s):
https://github.com/LibertyDSNP/spec/issues/169

Google Doc with Notes / Discussion:
https://docs.google.com/document/d/1ON2cJRvZYFWmTD6VzLXNnt2B4CijrdYfWap9JH6uw9E/edit

Solution
========
This proposal includes graph change types for blocking/unblocking.

Notes
========
It may be possible to accommodate blocking at the level of ActivityContent. This would strike a convenient middle ground of allowing spec implementers to offer a blocking feature to their users without burdening governance with the task of enforcing blocking behavior throughout the network.

Including this feature in the DSNP spec may also make the protocol bulkier than necessary. We should bide our time and wait to see how a live social network might want to use this feature with an active user base.


saraswatpuneet (dismissed):

saraswatpuneet (approved):

wilwade (changes requested):
@dopatraman Could you fill out the DIP issue with a bit more detail? I'll add some extra questions and notes in the comments over there.

wilwade:
Closing for now. We are going to think more about this with respect to private graph

PR #172, opened 2022-04-14
Chore/sync activity content spec/1

mehtaishita:
Problem
=======
Make changes to address inconsistencies between ActivityContent package and Spec

https://github.com/LibertyDSNP/activity-content/issues/1

Solution
========
Making changes to the Activity Content portion of the spec to better clarify required values and parameter names

Change summary:
--------------
- [x] Tidy, well formulated change
- [x] Rename URL properties for Associated types
- [x] Update respective examples to match
- [x] Amended Associated Type: Hash example

Steps to Verify:
----------------
1. Build spec
2. Verify changes are consistent across various attachment types - Audio, Video, Image


mehtaishita:
From description of the ticket:
_For ActvityContentHash - in spec value is required. In SDK we have dynamic values for key. - go with spec

For ActivityContentLocation - spec says name is required -
SDK has name optional - so need to resolve - required_

Both of these are required in the sdk so there are no changes to make there.

In the description, we also decided to not pluralize parameter names, so I have not renamed Video Link: Hash to Hashes. Please verify that decision

pages/DSNP/Types/GraphChange.md
```diff
@@ -31,6 +31,8 @@ Different change types have different meanings
 | ----- |----- | ----------- |
 | 0 | Unfollow | Remove a Follow relationship |
 | 1 | Follow | Create a Follow relationship |
+| 2 | Block | Create a Block relationship |
```
shannonwells, line 34: We haven't yet decided to add Block/Unblock to the protocol. I think we are _mostly_ against it being in the protocol but it's still not decided.

shannonwells (dismissed):
Looks good, just need to remove the block/unblock stuff.  It's out of scope for the ticket anyway.

wilwade (changes requested):
I think we might have gotten the direction flipped on this PR. I think the changes were needed in https://github.com/LibertyDSNP/activity-content not the spec.

mehtaishita:
As in change the variable names in activity-content to href?

wilwade:
> As in change the variable names in activity-content to href?

Right. The spec is correct, the code in activity content is wrong.

mehtaishita:
I have reverted the renaming parameters for Associated Attachment types

pages/ActivityContent/Associated/Hash.md
```diff
@@ -26,6 +26,10 @@ AT LEAST ONE hash in the array MUST be one of the [supported algorithms](#suppor
     {
       "algorithm": "keccak256",
       "value": "0x1234567890ABCDEF0123456789ABCDEF0123456789ABCDEF0123456789ABCDEF"
+    },
+    {
+      "algorithm": "Ripemd256",
+      "value": "96a9209006748794193d1811ef2dd5f447782b8b1635841165bc031bb3db64da"
```
wilwade, line 32: 0x

wilwade (approved):

Issue #149, opened 2022-03-24
DIP-149 Reaction Retraction


wilwade:
# Abstract

Update the Reaction Announcement to have a revoke bit flag and remove it from being targeted by Tombstones.

# Motivation

This comes out of discussion of DIP-#145.
DIP-145 forces a change in Tombstones, but that change is made more difficult by reactions being tombstoned.
When looking at options for tombstone changes, it was discovered that altering reaction retraction may be worthwhile either way.

# Specification Pull Request

Current change pull request: #173 

# Rationale

One of the categories of announcements we have is this: Ones that have a link to external content and ones that do not.
Reactions are the only announcement type that does NOT link to external content that can be Tombstoned.

Second, this change should increase the resolution speed of Reactions as one does not need to check each reaction for a tombstone.
Instead all reactions and retractions are associated at the same level.

### Alternatives:
- Keep using Tombstones
  - See DIP-145 discussion
- An announcement type specifically for reaction retraction
  - That's a lot of additional searching
- Not allowing retraction
  - We want users to be able to make mistakes because we do.

## Resolution Pattern Notes

How do we want to resolve if someone has or has not reacted?

Assumption: No user may use the same reaction more than once per item to react to.

### Accepted: Option 1: Time Based State Transition

This follows the pattern for graph. Merely play through the reaction announcements in order ignoring duplicates and assuming that one cannot retract before one reacts.

This could also be referred to as "Final Answer Wins" as the result is the same.

So a user reacting:
-  `+` => Yes
-  `+ -` => No
-  `+ - + -` => No
-  `+ - - +` => Yes
-  `+ - - -` => No
-  `- - - +` => Yes
-  `+ + + -` => No

### Rejected: Option 2: Time Independent Counting

This is a simpler resolution in that it is not time dependent.
Merely count the number of reactions and the number of retractions per user per reaction. If the reaction count is greater than the retraction count, then it is a (single) reaction.

So a user reacting:
-  `+` => Yes
-  `+ -` => No
-  `+ - + -` => No
-  `+ - - +` => **No**
-  `+ - - -` => No
-  `- - - +` => **No**
-  `+ + + -` => **Yes**

# Backwards Compatibility

This is a breaking change for the reaction announcement type.

We could just deprecate the current Announcement Type Enum and create a new one, but we have a collection of breaking changes for v1.1 (DIP-145 for example), and this could be included with those.

# Reference Implementation and/or Tests

[...]

PR #173, opened 2022-04-14
Reaction retractions for DIP-149

shannonwells:
## Problem
Specify #149 Reaction Retraction

with @mehtaishita 

## Solution
Somewhat of a different solution than proposed:

Have you ever commented, "I wish I could ❤️ x 1000 this post" ?  This spec change allows for that.

Add an "apply" field to a Reaction.  The "apply" field has a numerical value, which is an unsigned, 8-bit integer, indicating whether to apply emoji and how "strong" of a reaction it is.

If the value is 0, the reaction is retracted.  If the value is N &gt; 0, then the reaction is applied, and its "strength" is N.  Using a non-zero, positive number to indicate strength opens up more uses for the Reaction message type besides a single emoji, such as
- surveys  
    "disagree strongly" &rarr; "agree strongly" = 1 &rarr; 5 stars
    each emoji reaction could be connected to a different question.
- ratings - emoji strength would be used to compute ratings, including percentages.
- recommendation engines (@mehtaishita's idea) - with the number of values possible in a UTF-8, this value could be repurposed to indicate a micro-segment of people with a set of preferences (such as taste in movies or TV shows) and the strength would indicate a confidence rating.

Change summary:
---------------
- [x] Updated Spec Versions


pages/DSNP/Types/Reaction.md
```diff
@@ -41,6 +42,22 @@ None of the following should be considered valid:
 ```
 "F", ":custom-emoji:", "<custom-emoji>", "ᚱ", "ᘐ", "״"
 ```
+### apply
+- MUST be a UINT_8
+Indicates whether the emoji should be applied and if so, its "strength".
```
wilwade, line 47: I'm not sure how I feel about allowing for a "strength" value to be applied. I like the potential use cases, but I'm not sure what the unintended consequences of allowing it might be.

The 0-255 range might be too much. I wonder about just a 4 bit number? (0-15) Hmm... It feels like we are trying to smash two separate things together.

pages/DSNP/Types/Reaction.md
```diff
@@ -41,6 +42,22 @@ None of the following should be considered valid:
...
+- ratings
+- a range of responses, e.g. "strongly disagree" --> "strongly agree" = 1 --> 5 stars.
+- recommendation engines
+
+#### Apply Enums
+
+| Value | Name    | Description                        |
+|-------|---------|------------------------------------|
+| 0     | retract | Remove the referenced emoji        |
+| n     | apply   | Apply the referenced emoji N times |
```
wilwade, line 60:
```suggestion
| 1-255     | apply   | Apply the referenced emoji N times |
```

pages/DSNP/Types/Reaction.md
shannonwells: I feel like it leaves some flexibility without actually breaking what a "reaction" is.  0-255 I think is fine, 0-15 is practically fine too... let's keep it 0-255 as your suggestion for now

wilwade (approved):

enddynayn (approved):

Issue #165, opened 2022-04-06
DIP-165: Updating Announcements


dopatraman:
# Abstract
Users of mainstream social networks may expect to be able to edit content that they post. This behavior is possible because centralized networks place control of content in a single administrative body that owns the entire network. Distributed networks, however, do not delegate ownership to any one body. As a result, editing content after posting publicly poses a unique challenge.

Announcements published to a distributed network that implements DSNP are immutable. While underlying content can change, the announcement itself, once published, lives on the blockchain in perpetuity. Below are some ideas on how we can ensure that users have an experience consistent with what they are used to without compromising the fundamentals of a distributed social network.

A separate Announcement Type for Updates will allow users to update content via DSNP.

# Motivation
Users coming from established social networks, specifically centralized ones, will expect the ability to edit content. This announcement type will allow them to do that.

# Specification Pull Request
https://github.com/LibertyDSNP/spec/pull/174

# Rationale
An Update Announcement type could function very similarly to a Tombstone. Along with the hash and url of the updated content, the announcement could also contain the signature of the original announcement along with a new signature of the creator and an announcement type.

This would require editable announcements to have either a signature on them that edit announcements can reference, or a content hash that can serve as an identifier. I don’t believe using a content URI would be enough, since the content hosted at that URI could always change.

This announcement would serve as an indicator to the other nodes in the network that the content at the url has changed with the approval of the creator.

### Alternatives:
An alternative could be to issue a tombstone for the original broadcast/reply and then issue another announcement of the same type. The upside of this approach is that it would not require a new announcement type.

The tombstone would need to indicate the signature of the announcement being updated as well as the type of that announcement. A validating body (either client or other validator node) would need to make sure the fromId matches the fromId of the original announcement.

One downside, however, to the approach is that it involves two separate announcements to achieve the edit/update. More actions required by the client increases surface area for errors.

# Copyright

Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/).

PR #174, opened 2022-04-18
Update Announcement Type for DIP-165

dopatraman:
Problem
=======
This PR adds an Update Announcement type to the DSNP spec.

Link to GitHub Issue(s):
#165 

Solution
========
The Update Announcement has a contentHash that is a keccak-256 hash of the _new, updated_ content. It also has a URL that points to the new content.

Discussion
========
Open question: will we need to create a new URL each time we update content?

Diagram:
Screen Shot 2022-04-18 at 3 57 39 PM

One scenario is as follows: a user emits a Broadcast, creating new content. The Broadcast Announcement contains a hash of the content as well as a URL that points to that content.

Later on, the user decides that they want to update said content. They emit an Update Announcement that contains a new hash of the updated content but the same URL as the original broadcast. This means the previous Broadcast's content hash will no longer be valid, even if it was valid at the time it was created.

Does this matter?

The upside of this approach is that we will be able to host content once and then point updates to that single hosted copy. The downside is that previous content hashes will be invalidated each time an update is made.

-------

Consider another scenario: the user emits a Broadcast creating content as earlier and tries to update it. This time, however, a new URL is created that points to either a brand new copy of the updated content, or some sort of diff.

Diagram:
Screen Shot 2022-04-18 at 4 00 51 PM

The upside of this approach is not only that the all content has a traceable history, but also that previous versions can be validated. However, the downside is that we will wither have to copy content on each update or rely on clients to piece together partial diffs.

**I think that until we have a compelling reason not to, we should take the 2nd approach.**

### DIFF
--- a/.yaspellerrc.json
+++ b/.yaspellerrc.json
@@ -83,7 +83,6 @@
     "blocklists",
     "changeType",
     "contentHash",
-    "createdAt",
     "deadDropID",
     "dequeued",
     "deserialized",
--- a/.yaspellerrc.json
+++ b/.yaspellerrc.json
@@ -113,7 +113,7 @@
     "namespace",
     "objectId",
     "officiation",
-    "contentURI",
+    "targetContentHash",
     "parseable",
     "permissioned",
     "prepended",
--- a/.yaspellerrc.json
+++ b/.yaspellerrc.json
@@ -113,7 +113,7 @@
     "namespace",
     "objectId",
     "officiation",
-    "oldUrl",
+    "contentURI",
     "parseable",
     "permissioned",
     "prepended",
--- a/.yaspellerrc.json
+++ b/.yaspellerrc.json
@@ -83,6 +83,7 @@
     "blocklists",
     "changeType",
     "contentHash",
+    "createdAt",
     "deadDropID",
     "dequeued",
     "deserialized",
--- a/.yaspellerrc.json
+++ b/.yaspellerrc.json
@@ -113,6 +113,7 @@
     "namespace",
     "objectId",
     "officiation",
+    "oldUrl",
     "parseable",
     "permissioned",
     "prepended",
--- a/pages/DSNP/Announcements.md
+++ b/pages/DSNP/Announcements.md
@@ -37,7 +37,7 @@ Additional duplicate Announcements MUST be rejected or ignored.
 
 ## Reverting an Announcement
 
-Announcements may not be deleted, but some may be marked as invalid by using a [Tombstone Announcement](Types/Tombstone.md).
+Announcements may not be deleted, but some may be marked as invalid by using a [Tombstone Announcement](Types/Tombstone.md), or updated by using an [Update Announcement](Types/Update.md).
 For example, if a user creates a Reaction Announcement, they may remove that reaction by creating a Tombstone Announcement.
 
 ## Non-Normative
--- a/pages/DSNP/Announcements.md
+++ b/pages/DSNP/Announcements.md
@@ -22,6 +22,7 @@ Each Announcement has an enumerated type for use when separating out a stream of
 | 3 | [Reply](Types/Reply.md) | a public response to a Broadcast | YES | YES |
 | 4 | [Reaction](Types/Reaction.md) | a public visual reply to a Broadcast | no | no |
 | 5 | [Profile](Types/Profile.md) | a profile | YES | no |
+| 6 | [Update](Types/Update.md) | an update to content| YES | no |
 
 ## Duplicate Handling
 
@@ -38,7 +39,6 @@ Additional duplicate Announcements MUST be rejected or ignored.
 
 Announcements may not be deleted, but some may be marked as invalid by using a [Tombstone Announcement](Types/Tombstone.md).
 For example, if a user creates a Reaction Announcement, they may remove that reaction by creating a Tombstone Announcement.
-
 
 ## Non-Normative
 
--- a/pages/DSNP/Types/Update.md
+++ b/pages/DSNP/Types/Update.md
@@ -12,7 +12,7 @@ announced content.
 | fromId | id of the user creating the announcement | 64 bit unsigned integer | [decimal](../Serializations.md#decimal) | `UINT_64` | YES
 | contentHash | keccak-256 hash of updated content | 32 bytes | [hexadecimal](../Serializations.md#hexadecimal) | `BYTE_ARRAY` | YES
 | url | updated content URL | UTF-8 | [UTF-8](https://datatracker.ietf.org/doc/html/rfc3629) | `UTF8` | no
-| oldUrl | URL of old content | UTF-8 | [UTF-8](https://datatracker.ietf.org/doc/html/rfc3629) | `UTF8` | no
+| contentURI | URL of old content | UTF-8 | [UTF-8](https://datatracker.ietf.org/doc/html/rfc3629) | `UTF8` | no
 
 ## Field Requirements
 
@@ -39,7 +39,7 @@ announced content.
 - Resource MUST one of the supported [Activity Content](../../ActivityContent/Overview.md) Types
 - MUST use one of the supported URL Schemes
 
-### oldUrl
+### contentURI
 
 - All above points for urls apply
 - MUST only point to Broadcast or Reply Announcements
--- a/pages/DSNP/Types/Update.md
+++ b/pages/DSNP/Types/Update.md
@@ -12,6 +12,7 @@ announced content.
 | fromId | id of the user creating the announcement | 64 bit unsigned integer | [decimal](../Serializations.md#decimal) | `UINT_64` | YES
 | contentHash | keccak-256 hash of updated content | 32 bytes | [hexadecimal](../Serializations.md#hexadecimal) | `BYTE_ARRAY` | YES
 | url | updated content URL | UTF-8 | [UTF-8](https://datatracker.ietf.org/doc/html/rfc3629) | `UTF8` | no
+| oldUrl | URL of old content | UTF-8 | [UTF-8](https://datatracker.ietf.org/doc/html/rfc3629) | `UTF8` | no
 
 ## Field Requirements
 
@@ -37,3 +38,8 @@ announced content.
 - MUST NOT refer to localhost or any reserved IP addresses as defined in [RFC6890](https://datatracker.ietf.org/doc/html/rfc6890)
 - Resource MUST one of the supported [Activity Content](../../ActivityContent/Overview.md) Types
 - MUST use one of the supported URL Schemes
+
+### oldUrl
+
+- All above points for urls apply
+- MUST only point to Broadcast or Reply Announcements
--- a/pages/DSNP/Types/Update.md
+++ b/pages/DSNP/Types/Update.md
@@ -9,7 +9,6 @@ Updates should be ignored.
 | Field | Description | Data Type | Serialization | Parquet Type | Bloom Filter |
 | ----- | ----------- | --------- | ------------- | ------------ | ------------ |
 | announcementType | Announcement Type Enum (`6`) | enum | [decimal](../Serializations.md#decimal) | `INT32` | no |
-| createdAt | milliseconds since Unix epoch | 64 bit unsigned integer | [decimal](../Serializations.md#decimal) | `UINT_64` | no
 | fromId | id of the user creating the announcement | 64 bit unsigned integer | [decimal](../Serializations.md#decimal) | `UINT_64` | YES
 | contentHash | keccak-256 hash of updated content | 32 bytes | [hexadecimal](../Serializations.md#hexadecimal) | `BYTE_ARRAY` | YES
 | url | updated content URL | UTF-8 | [UTF-8](https://datatracker.ietf.org/doc/html/rfc3629) | `UTF8` | no
@@ -20,10 +19,6 @@ Updates should be ignored.
 ### announcementType
 
 - MUST be fixed to `6`
-
-### createdAt
-
-- MUST be set to the milliseconds since Unix epoch at time of signing
 
 ### fromId
 
--- a/pages/DSNP/Types/Update.md
+++ b/pages/DSNP/Types/Update.md
@@ -1,7 +1,8 @@
 # Update Announcement
 
 An Update Announcement is a way to note intent to update previously
-announced content.
+announced content. If the original Broadcast/Reply is Tombstoned, subsequent
+Updates should be ignored.
 
 ## Fields
 
--- a/pages/DSNP/Types/Update.md
+++ b/pages/DSNP/Types/Update.md
@@ -1,0 +1,45 @@
+# Update Announcement
+
+An Update Announcement is a way to note intent to update previously
+announced content.
+
+## Fields
+
+| Field | Description | Data Type | Serialization | Parquet Type | Bloom Filter |
+| ----- | ----------- | --------- | ------------- | ------------ | ------------ |
+| announcementType | Announcement Type Enum (`6`) | enum | [decimal](../Serializations.md#decimal) | `INT32` | no |
+| createdAt | milliseconds since Unix epoch | 64 bit unsigned integer | [decimal](../Serializations.md#decimal) | `UINT_64` | no
+| fromId | id of the user creating the announcement | 64 bit unsigned integer | [decimal](../Serializations.md#decimal) | `UINT_64` | YES
+| contentHash | keccak-256 hash of updated content | 32 bytes | [hexadecimal](../Serializations.md#hexadecimal) | `BYTE_ARRAY` | YES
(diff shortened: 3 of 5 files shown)


pages/DSNP/Types/Update.md
```diff
@@ -0,0 +1,57 @@
...
+
+- MUST be the [keccak-256 hash](https://keccak.team/files/Keccak-submission-3.pdf) of the bytes of the reference at the url
+
+### url
+
+- MUST NOT refer to localhost or any reserved IP addresses as defined in [RFC6890](https://datatracker.ietf.org/doc/html/rfc6890)
+- Resource MUST one of the supported [Activity Content](../../ActivityContent/Overview.md) Types
+- MUST use one of the supported URL Schemes
+
+#### Tombstone Allowed Announcement Types
```
shannonwells, line 42: Did you mean "Update Allowed Announcement Types"?

pages/DSNP/Types/Update.md
```diff
@@ -0,0 +1,57 @@
...
+| fromId | id of the user creating the announcement and tombstoned announcement | 64 bit unsigned integer | [decimal](../Serializations.md#decimal) | `UINT_64` | YES
+| contentHash | keccak-256 hash of updated content | 32 bytes | [hexadecimal](../Serializations.md#hexadecimal) | `BYTE_ARRAY` | YES
+| url | updated content URL | UTF-8 | [UTF-8](https://datatracker.ietf.org/doc/html/rfc3629) | `UTF8` | no
+| signature | creator signature | 65 bytes | [hexadecimal](../Serializations.md#hexadecimal) | `BYTE_ARRAY` | no
+
+## Field Requirements
+
+### announcementType
+
+- MUST be fixed to `0`
```
shannonwells, line 21: According to the row added for this, the announcementType is 6. Was this a missed edit?

pages/DSNP/Announcements.md
```diff
@@ -40,7 +40,6 @@ Additional duplicate Announcements MUST be rejected or ignored.
 Announcements may not be deleted, but some may be marked as invalid by using a [Tombstone Announcement](Types/Tombstone.md).
```
shannonwells, line 40:
```suggestion
Announcements may not be deleted, but some may be marked as invalid by using a [Tombstone Announcement](Types/Tombstone.md), or updated by using an [Update Announcement](Types/Update.md).
```

shannonwells (commented):
Never mind re: DIP, I just forgot about what was in the PR description...

pages/DSNP/Types/Update.md
dopatraman: Yes, good catch 😬

dopatraman:
Suggestion: We can also name this announcement "Update Content", as it can only be used to update content.

pages/DSNP/Types/Update.md
```diff
@@ -0,0 +1,45 @@
...
+## Fields
+
+| Field | Description | Data Type | Serialization | Parquet Type | Bloom Filter |
+| ----- | ----------- | --------- | ------------- | ------------ | ------------ |
+| announcementType | Announcement Type Enum (`6`) | enum | [decimal](../Serializations.md#decimal) | `INT32` | no |
+| createdAt | milliseconds since Unix epoch | 64 bit unsigned integer | [decimal](../Serializations.md#decimal) | `UINT_64` | no
+| fromId | id of the user creating the announcement | 64 bit unsigned integer | [decimal](../Serializations.md#decimal) | `UINT_64` | YES
+| contentHash | keccak-256 hash of updated content | 32 bytes | [hexadecimal](../Serializations.md#hexadecimal) | `BYTE_ARRAY` | YES
+| url | updated content URL | UTF-8 | [UTF-8](https://datatracker.ietf.org/doc/html/rfc3629) | `UTF8` | no
+| signature | creator signature | 65 bytes | [hexadecimal](../Serializations.md#hexadecimal) | `BYTE_ARRAY` | no
```
aramikm, line 15: so this is missing one of the most important fields. Which announcement or content is getting updated?

pages/DSNP/Types/Update.md
```diff
@@ -0,0 +1,45 @@
+# Update Announcement
+
+An Update Announcement is a way to note intent to update previously
+announced content.
+
+## Fields
+
+| Field | Description | Data Type | Serialization | Parquet Type | Bloom Filter |
+| ----- | ----------- | --------- | ------------- | ------------ | ------------ |
```
aramikm, line 9: based on https://github.com/LibertyDSNP/spec/issues/145 we probably don't need the signature field

pages/DSNP/Types/Update.md
aramikm: This should not be `0`

pages/DSNP/Types/Update.md
dopatraman: So I think we could go one of two ways on this.

One is to have an identifier that points to a previous announcement. However, if that previous announcement is itself an Update, the client will have to look up (potentially multiple) other announcements. This is possible, but it could be expensive. Still possible, though.

Another option is to leave out the pointer altogether. We simply broadcast intent to update with a URL that points to the most recent content. Clients, after receiving an Update announcement, could simply update the URL / content. I decided to go with this approach here bc it seemed simpler.

Thoughts?

pages/DSNP/Types/Update.md
```diff
@@ -0,0 +1,45 @@
...
+### url
+
+- MUST NOT refer to localhost or any reserved IP addresses as defined in [RFC6890](https://datatracker.ietf.org/doc/html/rfc6890)
+- Resource MUST one of the supported [Activity Content](../../ActivityContent/Overview.md) Types
+- MUST use one of the supported URL Schemes
+
+### signature
+
+- MUST be an [Announcement Signature](../Signatures.md) over the all fields except this signature field
+- MUST be the signature of the `fromId` user
```
saraswatpuneet, line 45: I think we are not doing the signature field

pages/DSNP/Types/Update.md
saraswatpuneet: Do client needs to know edit history ? Will there be a use case for that, in that case we may link to previous one and follow the trail

dopatraman: Please see updates.

.yaspellerrc.json
```diff
@@ -114,6 +114,7 @@
     "namespace",
     "objectId",
     "officiation",
+    "oldUrl",
```
saraswatpuneet, line 117: If everyone agrees with calling it oldUri or something more dsnpish wording rest looks good

dopatraman:
@saraswatpuneet any thoughts on the question of whether or not we should generate a new URI / copy of content?

saraswatpuneet:
@dopatraman  I think since we are linking back to old uri we might generate a new one which in case get updated again will point to previous one as you mentioned to have a have a consistent flow of edits/update but @wilwade's insight in this will be useful

aramikm:
My personal choice would be to use 2nd approach meaning creating a new update message that points to original broadcast or reply Message (maybe using content URI or another announcement identifier) and banning creation of any Update message for another update messages. (Kinda like not allowing to tombstone another tombstone message). So we will always apply and use content of the latest update message.

first operation -> adding a Broadcast (B)
update 1 -> adding an Update (U1) which is pointing to B
update 2 -> adding another update (U2) which is also pointing to B (if we always keep the latest copy of the content we do not need any chaining but if we decide to do diffs then we would need to do chaining)

saraswatpuneet:
@aramikm  that definitely is an approach but then aren't we assuming that only one update is allowed is enforcing a rule but  if we be flexible for updates l, services would pay for it and they can do this custom restrictions on their end giving them flexibility to whether allow one two or n updates 


Thought? Though if we want to restrict we need to have a rationale but I am ok with either or

aramikm:
> @aramikm that definitely is an approach but then aren't we assuming that only one update is allowed is enforcing a rule but if we be flexible for updates l, services would pay for it and they can do this custom restrictions on their end giving them flexibility to whether allow one two or n updates
> 
> Thought? Though if we want to restrict we need to have a rationale but I am ok with either or

So my description probably is not clear. Based on that approach that I was thinking you can have unlimited updates but any update will just point to the original announcement instead of pointing to it's previous update. The reason is that if you always keep the latest content you don't need to follow the chain of updates.

dopatraman:
@saraswatpuneet i see what you're saying -- that by only allowing updates to be made to original content we are indirectly creating a rule that there can be only one update. However, it may be possible for clients to:

1. aggregate updates by announcement URI
2. apply the updates in order (ordered by timestamp maybe?) to reach the final state of the content

that way, we can still prevent users from making an update to an update without indirectly limiting the number of updates that can be made.

Thoughts?

shannonwells (approved):
It looks good to me as is. I agree with the idea of always pointing to the original post as the oldUri.  We shouldn't be worrying about edit history since we really only care that the end user is displaying the content they want to display. If someone cares about edit history, a social network is not a good platform for that.

saraswatpuneet (approved):

.yaspellerrc.json
aramikm: I prefer something like `target`. I don't think this should be a url, to me `contentUri` seems more compatible with existing spec.

pages/DSNP/Announcements.md
```diff
@@ -22,7 +22,7 @@ Each Announcement has a enumerated type for use when separating out a stream of
 | 3 | [Reply](Types/Reply.md) | a public response to a Broadcast | YES | YES |
 | 4 | [Reaction](Types/Reaction.md) | a public visual reply to a Broadcast | no | YES |
 | 5 | [Profile](Types/Profile.md) | a profile | YES | no |
-
+| 6 | [Update](Types/Update.md) | an update to content| YES | YES |
```
wilwade, line 25: Should Updates be tombstonable? Or is it just replaced with another edit?

pages/DSNP/Types/Update.md
```diff
@@ -0,0 +1,45 @@
...
+## Fields
+
+| Field | Description | Data Type | Serialization | Parquet Type | Bloom Filter |
+| ----- | ----------- | --------- | ------------- | ------------ | ------------ |
+| announcementType | Announcement Type Enum (`6`) | enum | [decimal](../Serializations.md#decimal) | `INT32` | no |
+| createdAt | milliseconds since Unix epoch | 64 bit unsigned integer | [decimal](../Serializations.md#decimal) | `UINT_64` | no
+| fromId | id of the user creating the announcement | 64 bit unsigned integer | [decimal](../Serializations.md#decimal) | `UINT_64` | YES
+| contentHash | keccak-256 hash of updated content | 32 bytes | [hexadecimal](../Serializations.md#hexadecimal) | `BYTE_ARRAY` | YES
+| url | updated content URL | UTF-8 | [UTF-8](https://datatracker.ietf.org/doc/html/rfc3629) | `UTF8` | no
+| oldUrl | URL of old content | UTF-8 | [UTF-8](https://datatracker.ietf.org/doc/html/rfc3629) | `UTF8` | no
```
wilwade, line 15: - Shouldn't this be the old content hash not the old URL?
- I would also expect it would be in the bloom filter as people would be searching for updates to a particular hash from a given user.

wilwade (changes requested):
Two blocking questions.

pages/DSNP/Announcements.md
dopatraman: I can see a scenario where a user might want to Tombstone an update as a way of saying "I'd like to undo this". I think it should be possible.

pages/DSNP/Types/Update.md
dopatraman: My understanding is that two users could, in theory, post the same content and so a contentHash by itself might not be enough to point to the target announcement. But the URI, again if I'm correct, would contain not only the content hash but a user's identifier as well, allowing us to precisely make an update.

But I realized in the process of writing this reply that if we assume that Updates can only be made to posts that have the same identifier as the `fromId`, then duplicate content hashes aren't really a concern. Can anyone think of a scenario where a user other than the original author would need to edit content?

pages/DSNP/Announcements.md
wilwade: That's a nested search however. Whereas a second `Update` announcement would be found in the first search and performs the same result right?

pages/DSNP/Types/Update.md
wilwade: I cannot think of a case where any other `fromId` should be allowed to edit a post. Even if others might be moderating, that would not be an edit.

pages/DSNP/Announcements.md
dopatraman: Just to be sure, what do we mean by "nested search"? If you mean that a Tombstone will force a client/listening node to do a lookup for the content to nullify, wouldn't that be true for another Update as well?

Another question -- is it possible that a user would want to nullify an Update as opposed to editing it with another Announcement? If both use cases are valid, my opinion is that we should allow it at the spec level and assume that developers implementing the spec will handle the exact implementation. Thoughts?

pages/DSNP/Types/Update.md
dopatraman: Makes sense. One thing that jumps out at me about an implementation that relies on the`fromId` is a potential N+1 (we'll have to fetch the user's announcements and then lookup the contenthash). The nice thing about using the URI is it will be a single lookup. Does that sounds correct to you?

pages/DSNP/Announcements.md
wilwade: Notes from the weekly spec discussion 2022-04-21

- Updates should NOT be able to be Tombstoned, just the original.
- If the original is Tombstoned, Update Announcements that target it are ignored

```suggestion
| 6 | [Update](Types/Update.md) | an update to content| YES | no |
```

That suggestion brings up the question of if Update Announcements have a DSNP Content URI. This line says yes, but I could see how we actually don't want it to be considered a new DSNP Content URI. Thoughts?

pages/DSNP/Types/Update.md
```diff
@@ -0,0 +1,44 @@
...
+
+| Field | Description | Data Type | Serialization | Parquet Type | Bloom Filter |
+| ----- | ----------- | --------- | ------------- | ------------ | ------------ |
+| announcementType | Announcement Type Enum (`6`) | enum | [decimal](../Serializations.md#decimal) | `INT32` | no |
+| createdAt | milliseconds since Unix epoch | 64 bit unsigned integer | [decimal](../Serializations.md#decimal) | `UINT_64` | no
+| fromId | id of the user creating the announcement | 64 bit unsigned integer | [decimal](../Serializations.md#decimal) | `UINT_64` | YES
+| contentHash | keccak-256 hash of updated content | 32 bytes | [hexadecimal](../Serializations.md#hexadecimal) | `BYTE_ARRAY` | YES
+| url | updated content URL | UTF-8 | [UTF-8](https://datatracker.ietf.org/doc/html/rfc3629) | `UTF8` | no
+| targetContentHash | keccak-256 hash of target content | 32 bytes | [hexadecimal](../Serializations.md#hexadecimal) | `BYTE_ARRAY` | YES
+
```
wilwade, line 16: @dopatraman 

I know we are getting rid of `signature` with DIP-#145, but do we need it until that is completed to merge this?

pages/DSNP/Announcements.md
dopatraman: Noting for posterity that we decided _not_ to allow tombstoning updates.

pages/DSNP/Types/Update.md
dopatraman: Good question. Since the goal is to get rid of signatures, I'm okay with merging this before 145. I'm curious what others think.

wilwade (changes requested):
Don't forget to remove createdAt now that DIP-145 is merged.

pages/DSNP/Types/Update.md
```diff
@@ -0,0 +1,39 @@
...
+### url
+
+- MUST NOT refer to localhost or any reserved IP addresses as defined in [RFC6890](https://datatracker.ietf.org/doc/html/rfc6890)
+- Resource MUST one of the supported [Activity Content](../../ActivityContent/Overview.md) Types
+- MUST use one of the supported URL Schemes
+
+### targetContentHash
+
+- MUST be the [keccak-256 hash](https://keccak.team/files/Keccak-submission-3.pdf) of the bytes of the reference at the url
```
wilwade, line 39: @dopatraman As I re-read this one last time, this line doesn't make sense. Perhaps

```suggestion
- MUST be the `contentHash` of an allowed Announcement type with the same `fromId` as the Update Announcement

#### Update Allowed Announcement Types

| Value | Name |
|------ | ---- |
| 2 | [Broadcast](../Types/Broadcast.md) |
| 3 | [Reply](../Types/Reply.md) |
```

wilwade (changes requested):
Found one more issue in the final read through.

wilwade (approved):

PR #175, opened 2022-04-19
First batch of copy edits

dopatraman:
Problem
=======
This PR makes the first batch of copy edits to the Spec Website.


saraswatpuneet:
@dopatraman  is it good to check in, you have approvals

PR #196, opened 2022-05-20
Switch to Raleway Font

wilwade:
Problem
=======

Switching to a new Font to match other coming changes to dsnp.org

Solution
========

- Stopped copying the fonts in from the theme
- Added the new font and the one we wanted to keep
- Update "logo" with just the DSNP text
- Set the sidebar to open at 800px instead of 1080

Screenshots (optional):
-----------------------
Light:
image

Dark:
image


enddynayn (approved):
Beautiful!

PR #205, opened 2022-09-14
DSNP on Frequency: Validation

### LINKED ISSUE #189: [DSNP] Frequency Batch Validation
Should be merged into
- `dsnp-on-mrc` Branch
- For PR: https://github.com/LibertyDSNP/spec/pull/197
- Fills out `pages/MRC/Validation.md`
- Effectively the version of https://github.com/LibertyDSNP/spec/blob/main/pages/Ethereum/Validation.md for Frequency

**Batch Publication Validation**

As a user, I need to know how to confirm the downloaded batch file is valid.

**Announcement Validation**

As a user, I need to know how to confirm the announcement in a given batch is valid.

**Announcement Duplicates**

As a user, I need to know how to handle duplicated announcements in a given batch or across batches.

wilwade:
Problem
=======

DSNP on Frequency needs to know how to validate announcements

Link to GitHub Issue(s): Closes #189 

Content dependent on changes pending: https://github.com/LibertyDSNP/frequency/issues/430

Screenshots (optional):
-----------------------
image

### DIFF
--- a/pages/Frequency/Validation.md
+++ b/pages/Frequency/Validation.md
@@ -1,16 +1,53 @@
 # Announcement Validation
 
-TODO: https://github.com/LibertyDSNP/spec/issues/189
+DSNP Announcements are validated differently depending on the type of Announcement.
+Batched Announcements are validated at read time.
+Non-batched Announcements (example: Graph Change), are validated at write time and do not need to be re-validated at read time.
 
-...
+## On-chain Announcements
 
-## Batch Publication Validation
+Creation of an Announcement as an on-chain Frequency Message requires that the delegation be validated before adding the message to the chain.
+Thus, any Graph Change Announcements read from a trusted node, can be trusted to have been made by the DSNP identity or a delegate of the identity at the time the Frequency Message was sent.
 
-...
+### On-Chain Announcement Types
+- Graph Change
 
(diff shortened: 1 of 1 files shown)


pages/Frequency/Publishing.md
```diff
@@ -1,24 +1,56 @@
...
-...
+Frequency Messages are well ordered within a Schema
+
+1. Frequency: Block number ascending
+2. Frequency: Transaction index ascending
+3. DSNP Standard: Order Announcements in a Batch Publication File by row appearance order
+
+### Ordering Across Schemas
+
+Frequency does not provide complete order information for Messages across Schemas.
```
aramikm, line 45: As I mentioned in another PR, we have the total order  across different schemas in the same block. How much can it be trusted based on different race conditions is up to the clients.

shannonwells (approved):
> Batched Announcements are validated at read time. Non-batched Announcements (example: Graph Change), are validated at write time and do not need to be re-validated at read time.

Do we make the connection elsewhere that batched announcements are off chain and non-batched are on chain?  Since this is going into the implementation spec, I wonder if we should say something like "Batched Announcements, being off-chain..." etc. just to make it clear and in case someone jumped straight to this page.

PR #209, opened 2022-09-29
Updated reference to MSA ID from RPC to state query.

### LINKED ISSUE #208: Update reference to `msa_getMsaId` RPC Call
### Steps to reproduce
- Look at the mention of `msa_getMsaId` on https://github.com/LibertyDSNP/spec/blob/main/pages/Frequency/Identity.md?plain=1#L27
- See that it was removed in https://github.com/LibertyDSNP/frequency/issues/456

### Expected
- Should reference the proper state query instead

rlaferla:
Closes #208


wilwade (approved):
Thanks for the quick fix

PR #215, opened 2022-11-21
DSNP Operations

### LINKED ISSUE #214: Formalize Operation Specifications
## Operations
- [ ] Identity
    - [ ] Create
    - [ ] Delegation Upsert
    - [ ] User Directed Delegation Revocation
    - [ ] Delegated Party Directed Delegation Revocation
    - [ ] DSNP Id Retire
    - [ ] Add DSNP Id Control Public Key
    - [ ] Remove DSNP Id Control Public Key
- [ ] Announcements
    - [ ] Publishing
    - [ ] Verification

## Additional Notes
- May be able to split up some of the data definition from actions like verification and such.

wilwade:
Problem
=======
The specification lacks formal details on the operations a compliant implementation should provide.
Link to GitHub Issue(s): Closes #214 

Solution
========
This PR adds Operations tables to the Identity and Announcements pages (which are the two areas where operations have thus far been specified). For each operation, we provide a name, the principal(s) who will be invoking the operation, the inputs, and the expected outputs or state changes.

Note that this is not a spec version change, but rather a clarification of the existing spec.

Change summary:
---------------
- [ ] Updated Overview page
- [ ] Updated Identity page
- [ ] Updated Announcements page

Steps to Verify:
----------------
1. Review proposed text for accuracy and clarity.
1. Generate the website book (`npm run build`) and preview in browser to ensure there are no formatting issues.

Screenshots (optional):
-----------------------
N/A


pages/DSNP/Identity.md
```diff
@@ -26,6 +26,22 @@ the ability to delegate permission to others to announce content on the user's b
 * Announcements from a delegate MUST be able to be verified as to which delegate made the specific announcement.
 * Delegation revocation MUST NOT be retroactive.
 
+## Operations
+
+Each of the these MUST have verifiable approval of the acting
+principal(s) via signature proof or other system. For example, an
+implementation might use the public key of an asymmetric key pair as a
+control key, and provide a proof for each operation by producing a
+cryptographic signature of the inputs and some nonce value.
```
wilwade, line 35:
```suggestion
Each of the these MUST have verifiable approval of the acting principal(s) via signature proof or other system.
For example, an implementation might use the public key of an asymmetric key pair as a control key, and provide a proof for each operation by producing a cryptographic signature of the inputs and some nonce value.
```

pages/DSNP/Announcements.md
```diff
@@ -40,6 +40,16 @@ Additional duplicate Announcements MUST be rejected or ignored.
 Announcements may not be deleted, but some may be marked as invalid by using a [Tombstone Announcement](Types/Tombstone.md), or updated by using an [Update Announcement](Types/Update.md).
 For example, if a user creates a Reaction Announcement, they may remove that reaction by creating a Tombstone Announcement.
 
+## Operations
+
+Each of the these MUST have verifiable approval of the acting
+principal via signature proof or other system.
```
wilwade, line 46: Here and elsewhere, we have tried to stick with one sentence per line (any length) to make diffs easier. (Here and on `Identity.md`). I don't think we have a lint to enforce that however as it is hard to know what is a sentence.

```suggestion
Each of the these MUST have verifiable approval of the acting principal via signature proof or other system.
```

pages/DSNP/Identity.md
```diff
@@ -26,6 +26,22 @@ the ability to delegate permission to others to announce content on the user's b
...
+
+| Operation | Optional? | Principal(s) | Inputs | Outputs/State Changes |
+|---------- |---------- |------------- |------- |-------------- |
+| Create DSNP Identifier | no | User | None | New DSNP Identifier connected to User's Control Key |
```
wilwade, line 39: We don't call it a DSNP Delegation, should we shorten this to just Create Identifier?
```suggestion
| Create Identifier | no | User | None | New DSNP Identifier connected to User's Control Key |
```

wilwade: User's control key should be an input here right? Although perhaps that is implied by the User as Principal? We don't require it on `Add Control Key`, but the same could apply there.
```suggestion
| Create DSNP Identifier | no | User | Control Key, Proof of Ownership of Control Key | New DSNP Identifier connected to User's Control Key |
```

pages/DSNP/Identity.md
```diff
@@ -26,6 +26,22 @@ the ability to delegate permission to others to announce content on the user's b
...
+| Create or Update Delegation | no | User AND Delegate | User's DSNP Identifier, Delegate's DSNP Identifier, Set of Announcement Permissions | Proof of Delegation |
```
wilwade, line 40: I know we talked about this some and I think I was on the combined side, but I am still on the fence about if this should be one or two lines. Thinking now about the `DSNP on [Impl]` side where it could be separate and I'm thinking now we should always have a 1:m relationship with implementations. Implementations could have more than one per item in the Ops list, but it would never have to split the ops list for operations that do just half of what is specified. Thoughts?

```suggestion
| Create Delegation | no | User AND Delegate | User's DSNP Identifier, Delegate's DSNP Identifier, Set of Announcement Permissions | Proof of Delegation |
| Update Delegation | no | User AND Delegate | User's DSNP Identifier, Delegate's DSNP Identifier, Set of Announcement Permissions | Proof of Delegation with New Permissions |
```

wilwade (commented):
Few notes and discussion questions

pages/DSNP/Announcements.md
```diff
@@ -40,6 +40,16 @@ Additional duplicate Announcements MUST be rejected or ignored.
...
+
+| Operation | Optional? | Principal | Inputs | Outputs/State Changes |
+|---------- |---------- |---------- |------- |-------------- |
+| Publish Announcement | no | User OR Delegate | User's DSNP Identifier, Announcement Type, Announcement Data | Announcement published to the network |
+| Publish Batch | no | User OR Delegate | User or Delegate's DSNP Identifier, Announcement Type, [Batch Publication File](BatchPublications.md) | Batch announcement published to network |
```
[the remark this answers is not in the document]
wesbiggs: My thoughts here...

> Should publish batch be a thing at all?

I'd like to consider this as two questions:
a) Should batching be part of the DSNP spec (as opposed to an implementation-specific spec for Frequency)?
b) If batching _is_ part of the DSNP spec, should it be an optional operation?

If batching is not in scope for DSNP, then we should move all the details of Parquet file structure and so on into the Frequency spec instead. I was proceeding from a basis of trying to capture what is already there in the DSNP spec docs. But I agree it's worth asking the question.

It is true that any batch announcement can be disaggregated into a series of individual announcements. An operational distinction is that all announcements in a batch must be from DSNP users who share a delegation relationship with the principal invoking the batch publish operation. However, once this relationship has been verified for each announcement, the fact that the delegate was the principal does not need to be retained as part of the state, and the state changes are equivalent.

On the other hand, I think we need to provide some clarity for _read_ operations on DSNP. If we define an observer role as one that sees all state change activity, then we can define the state change event stream in an abstract form. If we say batches are an implementation detail, then an observer (i.e. a reader who is outside the boundaries of the implementation and relying on the spec operations it provides) can be defined to see individual announcement events only. If, however, we want to put the responsibility of reading batch files onto the observer, then the batch announcement becomes its own event on the event stream, and the format of the batch file needs to be specified on the DSNP side.

Working it through this way, I would like to propose the following:
1. We document a formal definition of a DSNP implementation as a distributed ledger, with an observable stream of records. Note that we don't need to specify the binary representation of these records (i.e. what it looks like on a blockchain) as long as we can specify the types of records and the data elements that must be observable. In other words, it must be possible to define an output translation function/layer that is able to produce a stream of normalized DSNP records for outside parties to interact with.
2. We specify individual announcements as part of the observable stream.
3. We DO NOT specify batches as part of the observable stream.
4. We move the definitions of batch file structure and handling out of the DSNP spec and make these the domain of the Frequency implementation team.

pages/DSNP/Identity.md
wesbiggs: From the point of view of an observable output record (the proof of delegation), I'm not sure there needs to be any distinction as long as the operation is idempotent. I see "Create or Update" (upsert, in developer-speak) a bit like a PUT operation in REST. Do you see a use case where an implementation would need to split it?

pages/DSNP/Identity.md
```diff
@@ -26,6 +26,22 @@ the ability to delegate permission to others to announce content on the user's b
...
+| Revoke Delegation | no | User OR Delegate | User's DSNP Identifier, Delegate's DSNP Identifier | Proof of Delegation Revocation |
+| Retire DSNP Identifier | no | User | User's DSNP Identifier | Allowed removal of all data sent from User's DSNP Identifier, Allowed removal of all Control Keys from DSNP Identifier, Allowed removal of all User's Delegations |
```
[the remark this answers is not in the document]
wesbiggs: So it looks like retiring an identifier is not actually described in the existing text of Identity.md, nor is it mentioned in the DSNP white paper. If it is a spec operation we should define it, which probably requires its own DIP and a spec version change.

Assuming we do define it, can we be more prescriptive here? If a DSNP Identifier is retired, and then there is a subsequent operation that attempts to use it, can we say it should not be allowed? (This is different from retaining data in storage, which is an implementation detail.)

pages/DSNP/Identity.md
wilwade: Split use case is when an implementation needs to split it. Blockchain actions that could be combined are often separated to cause less on-chain work (and thus cost less). So I think it likely that an implementation have create and update separate.

It is different in Frequency:

- Create: `grant_delegation`
- Update: `grant_schema_permissions` AND `revoke_schema_permissions`

pages/DSNP/Announcements.md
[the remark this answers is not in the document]
wilwade: Hmm... It feels odd for batching to be separate from DSNP as it is a core piece to making it work on any current scaled implementation. Or perhaps a better way to put it is that it doesn't feel right that it is part of the Frequency on DSNP spec. I suppose it could be a "separate" spec (like Activity Content).

The responsibility of reading batch files is on the observer. We can try to say that it is merely part of the implementation, but that is more arbitrary than most of the lines. It also "removes" the value of using the bloom filters to skip processing batch files that are not important to your application.

The event stream in theory is nothing but Announcements and it is designed to be consumed in that way, but it does reduce the power of the Batches.

pages/DSNP/Identity.md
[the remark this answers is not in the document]
wilwade: Hmm.... Perhaps it is an optional spec piece? You are right about the data. That is a very implementation specific.

I would be ok with optional or removing this and making it just a part of the implementation spec.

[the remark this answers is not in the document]
shannonwells: I don't think this should be optional. Arguments in favor, in no particular order:

1. We talk and have written about the ability to delete/retire many times. Leaving it out of the specification implies the right to be forgotten is optional or even orthogonal to DSNP, which to me doesn't make sense, especially since we're about giving users control over their social identity. "[those] who can destroy a thing, control a thing", or so some Herbert dude once wrote.

2. If we specify Tombstones why would we not support retiring identifiers?

3. Why eliminate the 'D' from CRUD just because it's identifier? What purpose does it serve to specify creation, updates, reads, but deny deletion (retire)?

4. Specifications often evolve well beyond white papers, so I don't consider that a strong argument against retiring identifiers.

A counterargument occurred to me that we don't specify any other operations regarding legal compliance, however, this is really the only portion of the specification that really applies, and leaving it out would explicitly ignore such laws. 

> Assuming we do define it, can we be more prescriptive here? If a DSNP Identifier is retired, and then there is a subsequent operation that attempts to use it, can we say it should not be allowed?

I think this is the best route, specify what should happen when the various operations include a retired identifier.

pages/DSNP/Identity.md
shannonwells: Is there a way to specify create/update and also allow them to be a united operation so long as the behavior is as specified?

pages/DSNP/Announcements.md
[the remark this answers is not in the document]
shannonwells: Since Frequency does not actually batch certain types of messages, in a way it is optional in that the Frequency implementation uses schema which explicitly say what messages are batched and what aren't.

Maybe that part should be part of the spec, what _may_ be batched and what _must not_ be batched.  Does it make sense to say specifically, e.g., you can't take identifier operations and graph messages out of the main/shared DSNP message stream, because that, it seems to me, undermines important goals of the protocol, like discoverability and verifiable delegation**?

I don't see a batch file author DSNP identifier in the [batch publication spec](https://spec.dsnp.org/DSNP/BatchPublications.html) by the way, there is just a reference that it should be able to be verified.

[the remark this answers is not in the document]
wesbiggs: @shannonwells I would turn this on its head somewhat and say that *because* Frequency has made some design decisions for efficiency, it (the implementation) needed to make certain choices about which announcement types are batched or not.

I feel that as long as the announcement operations end up emitting records (events) that reference the announcements or batches and give users a way to retrieve them, discoverability should be OK.

Perhaps what we need to do is specify that certain announcement types must have their delegation verified by the implementation vs. by the reader.

pages/DSNP/Identity.md
wesbiggs: > We don't call it a DSNP Delegation, should we shorten this to just Create Identifier?

That certainly makes things more concise, and since that's the only type of identifier we're interested in, I think we can change "DSNP Identifier" to "Identifier" throughout the operation spec sections.

> User's control key should be an input here right? Although perhaps that is implied by the User as Principal? We don't require it on Add Control Key, but the same could apply there.

I agree, I think we can be explicit here that the control key (and proof) are parameters, since no user authentication context exists yet at the time of operation invocation.

pages/DSNP/Identity.md
wesbiggs: Let's imagine a hypothetical "translation layer" converting DSNP operations to Frequency RPC calls. As long as this translation layer has the ability to access current state data (i.e. what delegation exists already), it can translate a "Define Delegation" call into any of grant_delegation, grant_schema_permissions, and revoke_schema_permissions. So for simplicity's sake I think there's a path to leaving it as a single operation in this spec.

pages/DSNP/Identity.md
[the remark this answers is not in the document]
wesbiggs: @shannonwells I agree with your points. I think "Retire Identifier" should be a required (and well specified) operation. Since that is a spec change I will propose a new DIP.

pages/DSNP/Identity.md
```diff
@@ -19,13 +19,34 @@ A user's ownership of their identity is expressed via ownership and control of t
...
+
+| Operation | Optional? | Principal(s) | Inputs | State Change Record |
+|---------- |---------- |------------- |------- |-------------- |
+| Create Identifier | no | User | Control Key, Control Key Ownership Proof | User Id Creation Record { User's Identifier, Control Key } |
+| Retire Identifier | no | User | None | User Id Retirement Record { User's Identifier } |
+| Define Delegation | no | User AND Delegate | User's Identifier, Delegate's Identifier, Set of Allowed  [Announcement Types](Announcements.md#announcement-types) | Delegation Definition Record { User's Identifier, Delegate's Identifier, Set of Allowed Announcement Types } |
+| Revoke Delegation | no | User OR Delegate | User's Identifier, Delegate's Identifier | Delegation Revocation Record { User's Identifier, Delegate's Identifier } |
+| Add Control Key | YES | User | Control Key, Control Key Ownership Proof | Control Key Addition Record { User's Identifier, Control Key } |
+| Remove Control Key | YES | User | Control Key | Control Key Removal Record { User's Identifier, Control Key } |
```
wilwade, line 47: Suggestion Only (Matter of visual syntax)

I don't quite get the `{}` syntax at the end of the State Change Record column. While it is trying to say these are the data items for the record, it is hard to parse in the format. I wonder if we want a different table of records or place to define each record to more clearly?

Perhaps another option is a subtable? (a bit gross for sure).

Another option might be separating out all the operations into a separate page and making each row a section? That would give a greater space to Inputs and State Change Record.

pages/DSNP/Identity.md
```diff
@@ -19,13 +19,34 @@ A user's ownership of their identity is expressed via ownership and control of t
 Control entails the power to announce content associated with the identifier and
 the ability to delegate permission to others to announce content on the user's behalf.
 
+## Retirement
+
+A user may choose to retire their identity at any time.
+Once an identifier is retired, an implementation MAY remove all state data associated with that identifier, provided that an indication that the identifier is retired remains, so it may not be reused in the future.
+This means that all data previously sent from the identifier, the control keys associated with the identifier, and the delegations (see next section) associated with the identifier may be removed.
+
 ### Delegation
```
wilwade, line 28: This was originally a sub heading of `## Ownership`. So we either want to move Retirement below it, or raise Delegation to `##` instead of `###`.

pages/DSNP/Overview.md
```diff
@@ -1,10 +1,31 @@
...
+ 
+All state changes MUST be visible to all participants in a DSNP system.
+In practical terms, this means that all users of a given DSNP system exist within a single overarching social graph, and must be able to see and participate in all public discourse on the network.
+
+Finally, to ensure decentralization, a DSNP system MUST avoid having any single point of failure (many nodes) and MUST avoid having any single entity that can override its consensus mechanisms (many operators).
+
+### Implementation Requirements
+
+Compliant DSNP implementations MUST document how each of the required DSNP Operations (for a given version of the DSNP specification) can be invoked, including, for example, what wire protocols should be used for discovery, authentication and operation execution, and how inputs and outputs will be serialized. 
+A compliant implementation MUST also specify a mapping from its implementation-specific state change data (for example, the events emitted by a blockchain) to the DSNP Records that data represents.
```
wilwade, line 27: Optional nit to remove "also"
```suggestion
A compliant implementation MUST specify a mapping from its implementation-specific state change data (for example, the events emitted by a blockchain) to the DSNP Records that data represents.
```

pages/DSNP/Identity.md
wesbiggs: Yes, I agree the format is a bit lacking.

How about this:
* A page listing all Operations
* A page listing all Record types
* A simple list of Operations (simple names with links) on Identity.md and Announcements.md

This provides an easier format (a checklist of sorts) for implementations to refer to and use as a template.

wesbiggs: OK I've updated this, let me know what you think.

pages/DSNP/Identity.md
wesbiggs: Thanks, will fix.

pages/DSNP/Overview.md
wesbiggs: Sure.

pages/DSNP/Operations.md
```diff
@@ -0,0 +1,35 @@
...
+
+Each invocation of a DSNP Operation MUST have verifiable approval of the acting principal(s) via a Control Key Ownership Proof.
+The precise data type and representation of both the Control Key and the Control Key Ownership Proof MUST be defined by each DSNP implementation.
+For example, an implementation might use the public key of an asymmetric key pair as a control key, and provide a proof for each operation by producing a cryptographic signature of the user's DSNP Identifier and some nonce value.
+
+Where operations are listed as using control keys or ownership proofs as input parameters, this indicates that they should be provided in addition to those providing invocation authentication.
+
+## Transaction Identifiers
+
+Each invocation of a DSNP Operation should be associated with a transaction identifier. The precise data type and usage pattern of a transaction identifier is left implementation-dependent.
```
aramikm, line 15: I think having transaction identifier as implementation-dependent will cause conflicts and eventually problematic for inter-operability between different implementations of DSNP

pages/DSNP/Announcements.md
```diff
@@ -40,6 +40,11 @@ Additional duplicate Announcements MUST be rejected or ignored.
 Announcements may not be deleted, but may be marked invalid by using a [Tombstone Announcement](Types/Tombstone.md), or updated by using an [Update Announcement](Types/Update.md).
 For example, if a user creates a Reaction Announcement, they may remove that reaction by creating a Tombstone Announcement.
```
jeanettedepatie, line 41: put a link for "Reaction Announcement"?

pages/DSNP/Identity.md
```diff
@@ -19,13 +19,27 @@ A user's ownership of their identity is expressed via ownership and control of t
 Control entails the power to announce content associated with the identifier and
 the ability to delegate permission to others to announce content on the user's behalf.
 
+### Retirement
+
+A user may choose to retire their identity at any time.
+Once an identifier is retired, an implementation MAY remove all state data associated with that identifier, provided that an indication that the identifier is retired remains, so it may not be reused in the future.
```
jeanettedepatie, line 25: Do we want to use an upper-case I on Identifiers as we use an upper-case A on Announcements?  The capitalization of Identifiers is not currently completely consistent within the document

pages/DSNP/Operations.md
```diff
@@ -0,0 +1,36 @@
...
+
+DSNP implementations perform well-defined DSNP Operations and generate DSNP State Change Records.
+
+## Control Keys and Proofs
+
+Each invocation of a DSNP Operation MUST have verifiable approval of the acting principal(s) via a Control Key Ownership Proof.
+The precise data type and representation of both the Control Key and the Control Key Ownership Proof MUST be defined by each DSNP implementation.
+For example, an implementation might use the public key of an asymmetric key pair as a control key, and provide a proof for each operation by producing a cryptographic signature of the user's DSNP Identifier and some nonce value.
+
+Where operations are listed as using control keys or ownership proofs as input parameters, this indicates that they should be provided in addition to those providing invocation authentication.
```
jeanettedepatie, line 11: It's not clear to me what "they" refers to in this sentence.

pages/DSNP/Operations.md
jeanettedepatie: Syntax issue: Uppercase T and I on Transaction Identifier?

pages/DSNP/Operations.md
```diff
@@ -0,0 +1,36 @@
...
+| <a id="revoke-delegation">Revoke Delegation</a> | no | User OR Delegate | User's Identifier, Delegate's Identifier | [Delegation Revocation Record](Records.md#delegation-revocation) |
+| <a id="add-control-key">Add Control Key</a> | YES | User | Control Key, Control Key Ownership Proof | [Control Key Addition Record](Records.md#control-key-addition) |
+| <a id="remove-control-key">Remove Control Key</a> | YES | User | Control Key | [Control Key Removal Record](Records.md#control-key-removal) |
+| <a id="publish-announcement">Publish Announcement</a> | no[^1] | User OR Delegate | [Announcement](Announcements.md) | [Announcement Published Record](Records.md#announcement-published) |
+| <a id="publish-batch">Publish Batch</a> | no[^1] | User OR Delegate | [Announcement Type](Announcements.md#announcement-types), [Batch Publication](BatchPublications.md) URL, Batch Publication Content Hash | [Batch Published Record](Records.md#batch-published) |
+
+[^1]: For each Announcement Type, an implementation may support one or both of these operations.
+    Implementations MUST document which of the operations is available for each announcement type.
```
jeanettedepatie, line 36: Syntax: Odd characters on screen after this sentence.

pages/DSNP/Overview.md
```diff
@@ -1,10 +1,32 @@
...
 
+## Overview
+
+### What is a DSNP System?
+
+A DSNP system is a state machine which generates an ongoing, publicly observable and verifiable stream of state change records in response to public input.
+This specification defines a set of operations (the DSNP [Operations](Operations.md)) that a compliant DSNP implementation MUST make available.
+For each DSNP Operation, the DSNP specification defines the data to be provided as input, the data that will be generated as output, and the state change records (the DSNP [State Change Records](Records.md)) that will be observable as a result of the Operation, depending on the input and prior state of the system.
+
+The nature of blockchain technology means that all DSNP Operations are potentially asynchronous; that is, DSNP State Change Records need not be created immediately upon invoking a DSNP Operation, and may appear in the stream of state changes at any time subsequently.
```
jeanettedepatie, line 16: or perhaps "at any subsequent time."

pages/DSNP/Overview.md
```diff
@@ -1,10 +1,32 @@
...
+A DSNP Operation can therefore only be confirmed to have completed successfully by observing the stream of newly generated DSNP Records.
+
```
jeanettedepatie, line 18: newly-generated (needs hyphen)

pages/DSNP/Records.md
```diff
@@ -0,0 +1,120 @@
+# State Change Records
+
+State Change Records constitute the observable output of a DSNP system.
+Implementations MUST specify how applications can translate implementation-specific output into State Change Records.
+
+Each Record is associated with a transaction identifier, which allows [Operations](Operations.md) to be associated with their results asynchronously.
+
```
jeanettedepatie, line 7: "Transaction Identifier" uppercase?

jeanettedepatie: perhaps "to be asynchronously associated with their results".

pages/DSNP/Operations.md
wesbiggs: > I think having transaction identifier as implementation-dependent will cause conflicts and eventually problematic for inter-operability between different implementations of DSNP

@aramikm I don't think it's much different from saying the control key algorithm is implementation-dependent. That said, if you think we can just say it will be a unique string, I'm fine with that. Let me know if I'm not understanding your concern here though.

pages/DSNP/Identity.md
wesbiggs: I like that suggestion. I'll change all usage in that document.

pages/DSNP/Operations.md
wesbiggs: Will fix.

pages/DSNP/Operations.md
wesbiggs: It looks like this syntax for footnotes is not completely standard for markdown tools. I'm going to change it to an asterisk instead.

wesbiggs:
Incorporated suggestions from @jeanettedepatie.

pages/DSNP/Operations.md
aramikm: I think it would be good to have that uniqueness requirement.
An example of my concern:
We have two DSNP implementations of A and B. Both of them decide to use incremental u64 as a transaction identifier. While trying to Interoperate between each other we will have 2 different transactions for each ID.

I think one way to avoid this scenario is to prefix the implementation specific name into transaction identifier. For following example that might looks like 'A://1', 'B://1', 'A://2', 'B://2', ....

shannonwells: > I think one way to avoid this scenario is to prefix the implementation specific name into transaction identifier. For mentioned example that might looks like 'A://1', 'B://1', 'A://2', 'B://2', ....

Could we use DID spec (or something like it, or something related) for that somehow?

wesbiggs: @aramikm Good idea... A prefix will give us a namespacing option for future interoperability. FWIW we also need this unique name to be incorporated into DSNP Content URLs to enable interoperability. The simplest thing to do in the context of this PR is to say that "dsnp://" is the prefix, because the current spec does not address interoperability. But I'd like to see interoperability (at least _potential_ interoperability) broken out into a separate DIP. This might entail Frequency changing to use something like "frqcy://" for its internal URLs/IRIs. To your point @shannonwells this implementation-specific prefix would need to be incorporated in the DID method. Options are "did:frqcy:123" or something like "did:dsnp:frqcy/123". The latter would in some respect require (or allow, depending on your point of view) a single "dsnp" DID method resolver that DSNP.org would maintain and allow the merging in of compliant implementations.

wesbiggs: Actually for now we can just ensure that Tx IDs are unique within an implementation, and then we can specify how they may be referenced from other implementations when we address interoperability (soon!).

shannonwells (approved):

pages/DSNP/Identity.md
```diff
@@ -6,18 +6,31 @@ This specification is intended to cover the concept of identity and how identity
...
+Optionally, an implementation MAY allow the user to add and remove additional keys.
+An implementation MUST NOT allow the user to remove a control key if doing so would result in the user no longer being able to assert control.
+
+#### Key Types
+
+The following Key Types are used in DSNP Operations:
+
+| Key Type | Serialized form (one byte) |
+| --- | --- |
+| Control Key | 0x00 |
```
shannonwells, line 33: Since Control keys (AccountId) in the Frequency implementation are 4 bytes, I'm not sure what this is saying.

pages/DSNP/Operations.md
```diff
@@ -0,0 +1,38 @@
+# Operations
+
+DSNP implementations perform well-defined DSNP Operations and generate DSNP State Change Records.
+
+## Control Keys and Proofs
+
+Each invocation of a DSNP Operation MUST have verifiable approval of the acting principal(s) via a Control Key Ownership Proof.
+The precise data type and representation of both the Control Key and the Control Key Ownership Proof MUST be defined by each DSNP implementation.
+For example, an implementation might use the public key of an asymmetric key pair as a control key, and provide a proof for each operation by producing a cryptographic signature of the user's DSNP Identifier and some nonce value.
```
shannonwells, line 9: Control keys also must be unique to an implementation - is that specified elsewhere? Could we say it must be an implementation-unique public key of a cryptographic key pair of at least N bits?

pages/DSNP/Overview.md
```diff
@@ -1,10 +1,32 @@
...
+This specification defines a set of operations (the DSNP [Operations](Operations.md)) that a compliant DSNP implementation MUST make available.
+For each DSNP Operation, the DSNP specification defines the data to be provided as input, the data that will be generated as output, and the state change records (the DSNP [State Change Records](Records.md)) that will be observable as a result of the Operation, depending on the input and prior state of the system.
+
+The nature of blockchain technology means that all DSNP Operations are potentially asynchronous; that is, DSNP State Change Records need not be created immediately upon invoking a DSNP Operation, and may appear in the stream of state changes at any subsequent time.
+A DSNP Operation can therefore only be confirmed to have completed successfully by observing the stream of newly-generated DSNP Records.
+ 
+All state changes MUST be visible to all participants in a DSNP system.
+In practical terms, this means that all users of a given DSNP system exist within a single overarching social graph, and must be able to see and participate in all public discourse on the network.
```
shannonwells, line 20: So we are explicitly saying that you cannot be a partial participant in a DSNP system and be compliant, i.e. you can't just post GraphChange messages and keep the rest "to yourself," away from the implementation?

pages/DSNP/Identity.md
```diff
@@ -6,18 +6,31 @@ This specification is intended to cover the concept of identity and how identity
...
+* An Identifier MUST default to being disconnected from a real-world identity.
+* An Identifier MUST be unique to the implementation.
 
 ## Ownership
 
-A user's ownership of their identity is expressed via ownership and control of their pseudo-anonymous identifier(s).
-Control entails the power to announce content associated with the identifier and
-the ability to delegate permission to others to announce content on the user's behalf.
+A user's ownership of their identity is expressed via ownership and control of their pseudo-anonymous Identifier(s).
+Control entails the power to announce content associated with the Identifier and the ability to delegate permission to others to announce content on the user's behalf.
```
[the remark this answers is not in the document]
wesbiggs: Control entails the power to perform DSNP Operations including publishing announcements that create, update and delete (tombstone) content associated with the Identifier, delegating these powers to others, managing keys associated with the Identifier, and retiring the Identifier [linked].

pages/DSNP/Operations.md
wesbiggs: Will add.

pages/DSNP/Announcements.md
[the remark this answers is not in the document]
wesbiggs: I'm going to move that last sentence to a discussion topic and resolve this thread.

pages/DSNP/Identity.md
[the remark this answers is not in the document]
wesbiggs: DIP-220 has been created for the ID retirement topic.

pages/DSNP/Identity.md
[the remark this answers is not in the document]
wesbiggs: The last bit on retiring the Identifier will move to the Retirement DIP-220 PR.

pages/DSNP/Operations.md
wesbiggs: Added a sentence in `Identity.md`.

wilwade (commented):
Apparently I created the initial PR @wesbiggs, but I approve it!

.yaspellerrc.json
```diff
@@ -52,7 +52,9 @@
     "Whitepaper",
     "allowlist[s]?",
     "announcementType",
-    "blockchain",
+    "archiveEntryID",
+    "asynchronously",  
+    "(a )?blockchain",
```
shannonwells, line 57: Non-blocking, but this line should be removed.
```suggestion
```

shannonwells (approved):

.yaspellerrc.json
wesbiggs: You would think so, right? I actually filed a bug for yaspeller on this: https://github.com/hcodes/yaspeller/issues/187 (unfortunately yaspeller as a project seems to have gone dark about the same time Russia invaded Ukraine).

wesbiggs (approved):
This is a Freaky Friday-style Wil approval.

Issue #228, opened 2023-01-30
DIP-228 Remove Graph Change Announcement


p5150j:
# Abstract

Proposes the removal of the Graph Change announcement type.

# Motivation

DSNP Users need to be able to manage their social graph in much more complex and controlled ways than what is possible with the current Graph Change announcement.

# Specification Pull Request

#230

# Rationale

A forthcoming DIP will propose that the social graph be considered mutable state and be managed via Replace User Data and Get User Data Operations. By making these DSNP Operations, implementations are empowered to optimize storage and make opinionated decisions about scalability and efficiency.

Additional points in favor of removal include the inability to manage private connections via a public announcement, and the lack of need for, and arguably, personal data leakage due to, publishing long-lived information about connections which may change.

# Backwards Compatibility

Not backward compatible. Implementations should ignore any previous Graph Change announcements and use the new User Data operations instead. However, at this early stage in the DSNP project, there should be few if any orphaned announcements for implementations to deal with.

# Reference Implementation and/or Tests

None at this time.

# Security Considerations

Removing Graph Change Announcements in favor of mutable state operations should give users better control over the social graph, which can only be a good thing for security.

# Dependencies

None.

# References

None at this time.

# Copyright

Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/).

wilwade:
Closed with #230

PR #230, opened 2023-01-31
DIP-228 Remove Graph Change Announcement

wesbiggs:
Problem
=======
Graph Change Announcement is too limited and will be replaced with mutable state management operations. See DIP #228.

Solution
========
This PR removes references to Graph Change Announcement. The new solution will be addressed in a separate PR.

Change summary:
---------------
- [ ] Removed Graph Change Announcement page
- [ ] Removed link from navigation section
- [ ] Removed from Announcement Type table
- [ ] Added to pending spec changes
- [ ] Removed a reference in Frequency docs that was now a broken link. (Note, however, that this PR does not attempt to update Frequency documentation.)

Steps to Verify:
----------------
1. Review content changes.
2. Generate site (`npm run build`) and check changed pages.


pages/Frequency/Publishing.md
```diff
@@ -1,14 +1,13 @@
 # Announcement Publishing
 
 On Frequency, [Announcements](../DSNP/Announcements.md) are mapped to Schemas which in turn publish Frequency Messages.
-Frequency Messages are either direct Graph Changes from a particular user, or a Batch Publication with a multitude of possible users.
+Frequency Messages are either mutable state changes from a particular user, or a Batch Publication with a multitude of possible users.
```
shannonwells, line 4: This implies an implementation cannot be DSNP compliant with if users can post their own announcements one at a time. This seems overspecified. What is the justification for preventing this (as opposed to simply supporting/allowing announcements in batches)?

wesbiggs: This page is for the Frequency spec, not DSNP. I just tweaked it so it didn't refer to Graph Change Announcements anymore. But this whole part of the spec site needs to be overhauled and updated outside the scope of this DIP.

shannonwells (approved):

pages/Frequency/Publishing.md
wilwade, line 4: Messages are never mutable state changes, but I get what is trying to happen here.

Perhaps:

```suggestion
On Frequency, [Announcements](../DSNP/Announcements.md) are mapped to Schemas which in turn publish Frequency Messages or state changes.
Frequency Messages can be from a particular user, or a Batch Publication with a multitude of possible users.
```

pages/Frequency/Publishing.md
```diff
@@ -1,14 +1,13 @@
...
 
 <!-- Links to https://libertydsnp.github.io/frequency should be updated with links to docs.frequency.xyz when able to be -->
 
 | Announcement Type Enum | Announcement | Batched | Schema Id Mainnet | Schema Id Rococo | Frequency Model Type | Frequency Payload Location |
 | --- | --- | --- | --- | --- | --- | --- |
 | 0 | [Tombstone](../DSNP/Types/Tombstone.md) | YES | TBD | 1 (_v1.1.0+_) | [`Parquet`](https://libertydsnp.github.io/frequency/common_primitives/schema/enum.ModelType.html#variant.Parquet) | [`IPFS`](https://libertydsnp.github.io/frequency/common_primitives/schema/enum.PayloadLocation.html#variant.IPFS) |
-| 1 | [Graph Change](../DSNP/Types/GraphChange.md) | no | TBD | 7 (_v1.1.0+_) | [`AvroBinary`](https://libertydsnp.github.io/frequency/common_primitives/schema/enum.ModelType.html#variant.AvroBinary) | [`OnChain`](https://libertydsnp.github.io/frequency/common_primitives/schema/enum.PayloadLocation.html#variant.OnChain) |
```
wilwade, line 11: Hmm... I know we likely will not publish Graph Change to mainnet Frequency, but feels odd just removing it.

Feels like it should have a version of `<= v1.1.0` or something. If we were to change another in the future, the older value (may) still matter to read historical data.

wilwade (changes requested):
Questions and one minor change to make for clarity.

wilwade (approved):
👍

Issue #227, opened 2023-01-30
DIP-227 User Data Operations


p5150j:
Create a DIP that describes the new Replace User Data and Get User Data operations, their parameters, and outputs, as well as the User Data Changed Record.

Issue #225, opened 2023-01-30
DIP-225 New primitives for social graph


p5150j:
Create a DIP for a new spec page on Social Graph, containing

* Overview of public and private graph
* Pseudonymous Recipient Identifier generation
* Relationship types
* Graph record schema (plaintext)

Separately, add the encoded Avro schema file for graph records to the repo.

PR #233, opened 2023-02-12
DIP-225/DIP-227 Graph and User Data Operations

wesbiggs:
Problem
=======
DSNP needs a way of specifying various user-related data that allows an implementation flexibility in storing social graph data while standardizing data formats and application layer data interaction semantics. GitHub issues: #225 #227 

Solution
========
The proposed user data spec strikes a compromise between the need for blockchains to segment data in optimal ways and the desire, from an application developer's point of view, to have data in an easy to work with format, whilst specifying details of schema, compression and encryption to ensure interoperability.

The user data types for the social graph that follow this approach are public follows, private follows, private connections, and private connection PRIds, and the spec describes how each is generated and interpreted.

Change summary:
---------------
- [ ] Added new page on User Data and included in navigation links
- [ ] Added items to Operations and Records tables
- [ ] Rewrote Graph page to define relevant user data types and the PRId algorithm

Steps to Verify:
----------------
1. Generate HTML with `npm run build`
2. Review spec site for content and formatting
1. Check for errors with `npm run lint`


pages/DSNP/UserData.md
```diff
@@ -0,0 +1,85 @@
...
+## User Data Types
+
+DSNP implementations MUST support the following user data types:
+
+| System Name | Version | Encryption Algorithm | Compression Codec | Avro Object Type |
+| --- | --- | --- | --- | --- |
+| `publicFollows` | 1.2 | NONE | [`brotli`](https://brotli.org) | [GraphEdge](Graph.md#edges) |
+| `privateFollows` | 1.2 | `curve25519xsalsa20poly1305` |  [`brotli`](https://brotli.org) | [GraphEdge](Graph.md#edges) |
+| `privateConnections` | 1.2 | `curve25519xsalsa20poly1305` | [`brotli`](https://brotli.org) | [GraphEdge](Graph.md#edges) |
+| `privateConnectionPRIds` | 1.2 | NONE | NONE | [PRId](Graph.md#pseudonymous-relationship-identifiers) |
```
aramikm, line 16: so no compression for PRIds based on experiments?

pages/DSNP/UserData.md
```diff
@@ -0,0 +1,85 @@
...
+
+## Replace User Data Operation
+
+The Replace User Data Operation takes the following parameters:
+
+* The DSNP User ID of the user who controls the data
+  * Implementations MUST ensure that the principal invoking this Operation is this user or a transparent chain of delegation from the user to the principal exists.
+* A map containing the set of data types to update, and the data chunks and entity tags for each included data type.
+
+The operation is idempotent.
```
aramikm, line 42: I don't think the update operation can be idempotent, since it would not be compatible with having `etags`.

pages/DSNP/UserData.md
```diff
@@ -0,0 +1,85 @@
...
+  * Note: While _writing_ user data is reserved for the user and any delegates, any authenticated principal can read any user's data (though it may be encrypted).
+* The user data types (by system name) that should be retrieved.
+
+The operation returns a mapping of user data type to data chunks and entity tags (the same general structure as the input data for Replace User Data), for each requested data type.
+If no chunks for a requested data type exist, an implementation MAY omit that data type from the response.
+
+To transform the public data from the output to Avro binary records, a consumer should apply the following algorithm to each data type included:
+* Determine the relevant encryption algorithm, compression codec, and object schema from the user data type and version noted.
+* For each chunk,
+  * If encryption is indicated, decrypt the chunk data using the user's secret key as in the [libsodium](https://doc.libsodium.org/public-key_cryptography/sealed_boxes) function `crypto_box_seal_open`.
```
aramikm, line 82: do we need to store any metadata along with encryption? same with Compression or since we are hardcoding this is protocol layer we are fine?

Is there any need to specify protocol version for any of these types?

pages/DSNP/Graph.md
```diff
@@ -2,29 +2,124 @@
...
 
-## Non-Normative
+Graph edges that originate from a user are stored as DSNP [User Data](UserData.md) associated with a given user.
 
-### Private Graph
+Each graph edge consists of the following data:
+* DSNP User Id (64-bit unsigned integer)
+  * To allow for optimal compression, User Ids are stored using the `long` type in Avro schema, which is a 64-bit _signed_ integer.
+    Care should be taken to ensure that User Id values greater than or equal to 2<sup>63</sup>, where used by an implementation, are converted correctly between signed and unsigned representations.
+* Timestamp when relationship was created (64-bit milliseconds since the [Unix epoch](https://en.wikipedia.org/wiki/Unix_time))
```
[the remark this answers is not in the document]
wesbiggs: My opinion is that it's useful to have, if only for the user themself to be able to track ("we've been friends since high school") or the app to use for curation ("happy friendiversary") etc. It also provides a useful way to give an ordering to the data that is a forcing function to put new relationships at the end, hopefully allowing for better transaction optimization.

As for granularity, I don't really see much actual use for more than a date, in which case 16 bits would be enough. If you add multiple friends on the same day, though, it does introduce some ambiguity from a consistent ordering perspective (you'd want to sort on date and then userId).

pages/DSNP/Graph.md
```diff
@@ -2,29 +2,124 @@
...
+Each public follow records a unidirectional relationship where one user (the controller of the graph) is interested in following the public speech of another user.
+When follows are public, they are visible to any other user or system, including the user being followed.
+
+The `publicFollows` user data type is composed of a set of GraphEdge objects.
+
+### Private Follows
+
+A private follow also records an interest in following the public speech of another user, but the data is stored in encrypted form, meaning the relationship is not visible to the followed user, or any other user, without access to the decryption key.
+
+The `privateFollows` user data type is composed of a set of GraphEdge objects.
```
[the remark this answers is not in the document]
wesbiggs: It's a compressed and encrypted set of GraphEdge objects. If that wasn't clear from context I'm happy to make it more explicit.

pages/DSNP/Graph.md
```diff
@@ -2,29 +2,124 @@
...
+
+1. Both Alice and Bob generate an asymmetric key pair, and each publishes a Public Key Announcement with their generated public key and a `keyType` value of `keyAgreement`.
+1. When Alice wants to interact with Bob, she looks up Bob's public key and performs a key exchange operation using her private key and Bob's public key, generating a shared secret `k` as in the libsodium function <code>crypto_box_beforenm(k, Bob<sub>public</sub>, Alice<sub>private</sub>)</code>.
+1. Alice derives a context-specific subkey <code>k<sub>PRId</sub></code>, using the shared secret `k` as the master key, as in the libsodium function `crypto_kdf_derive_from_key`, using `0` (zero) as the key identifier (see the [PRId Context](#contexts) table below) and the ASCII encoding of `"DSNPPRId"` as the 8-byte human-readable identifier.
+1. Alice concatenates her DSNP User Id and Bob's DSNP User Id to form a 16-byte message.
+1. Alice encrypts the message using the PRId key <code>k<sub>PRId</sub></code> and a nonce of all zeros, as in the libsodium function `crypto_secret_box_easy`.
```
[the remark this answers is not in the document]
wesbiggs: Yes. And good point, the same nonce shouldn't be reused with the same key. I need to change it so that the derived key is specific to the relationship target... I forgot about Bob using the same key/nonce. If the derived keys are different, a fixed nonce is OK because the key is only ever used for this one encryption.

pages/DSNP/UserData.md
wesbiggs: We need a decision on Bloom (or other type of) Filter because that seems like the only way to save significant space. Without that I feel like a <5% gain (with delta encoding) is not worth the burden it puts on rolling our own DSNP-specific compression codec.

pages/DSNP/UserData.md
wesbiggs: Good catch, will remove.

pages/DSNP/UserData.md
wesbiggs: Data type and version should be enough to perform a lookup on what compression/encryption is needed. But to that point I will make it clear that version is required to be noted for each data type.

pages/DSNP/Graph.md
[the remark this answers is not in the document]
wesbiggs: A couple of further notes on this:
* Avro doesn't have a 16-byte primitive type, so we would have to define serialization to `fixed(2)`.
* For reference, the millisecond-based approach with an Avro `long` will ZigZag encode to 48 bits for the foreseeable future, so it's not as bad as the full 64 (but still a lot more than 16!)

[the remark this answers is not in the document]
wilwade: I'm on the side of seconds over milliseconds for data that usually goes on-chain.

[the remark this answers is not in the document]
wilwade: Re: `long` Our announcements have a Parquet data type. Do we also want to have an avro type in there?

pages/DSNP/Graph.md
```diff
@@ -2,29 +2,124 @@
...
-The public graph is anchored in state transitions recorded in [Graph Change Announcements](Types/GraphChange.md).
-Playing all state transitions in [order](Announcements.md#ordering-announcements) will generate the current state of the public graph.
+## Edges
 
-## Non-Normative
+Graph edges that originate from a user are stored as DSNP [User Data](UserData.md) associated with a given user.
 
-### Private Graph
+Each graph edge consists of the following data:
+* DSNP User Id (64-bit unsigned integer)
```
wilwade, line 17: The signed/unsigned is strange here. It feels odd to have a uint64 everywhere but here.

pages/DSNP/Graph.md
```diff
@@ -2,29 +2,124 @@
...
+
+1. Both Alice and Bob generate an asymmetric key pair, and each publishes a Public Key Announcement with their generated public key and a `keyType` value of `keyAgreement`.
+1. When Alice wants to interact with Bob, she looks up Bob's public key and performs a key exchange operation using her private key and Bob's public key, generating a shared secret `k` as in the libsodium function <code>crypto_box_beforenm(k, Bob<sub>public</sub>, Alice<sub>private</sub>)</code>.
+1. Alice derives a context-specific subkey <code>k<sub>Bob</sub></code>, using the shared secret `k` as the master key, as in the libsodium function `crypto_kdf_derive_from_key`, using Bob's DSNP User Id as the 64-bit key identifier and the ASCII encoding of `"DSNPPRId"` as the 8-byte human-readable identifier.
+1. Alice concatenates her DSNP User Id and Bob's DSNP User Id to form a 16-byte message.
+1. Alice encrypts the message using the PRId key <code>k<sub>Bob</sub></code> and a nonce of the [PRId Context](#contexts) (all zeros for connections), as in the libsodium function `crypto_secret_box_easy`.
```
wilwade, line 105:
```suggestion
1. Alice encrypts the message using the PRId key <code>k<sub>Bob</sub></code> and a nonce of the [PRId Context](#contexts) (all zeros for connections), as in the libsodium function `crypto_secretbox_easy`.
```

pages/DSNP/UserData.md
```diff
@@ -0,0 +1,84 @@
+# User Data
+
+DSNP stores user-related data of well defined formats associated with a user's DSNP Identifier.
+Data may be public or stored in encrypted form.
+The [Avro](https://avro.apache.org) schema specification is used to define the binary representation of relevant data types.
```
wilwade, line 5: We should expand some here as to what user data is. Here's a shot at it, but it is just a start:

```suggestion
DSNP stores user-related data of well defined formats associated with a user's DSNP Identifier.
User-related data is for data about the user that is intended to be accessible across the network, but does not need to be widely announced and instead can be read just-in-time for a particular user.
Data may be public or stored in encrypted form.
The [Avro](https://avro.apache.org) schema specification is used to define the binary representation of relevant data types.
```

pages/DSNP/UserData.md
```diff
@@ -0,0 +1,84 @@
...
+
+The operation returns a mapping of user data type to data chunks and entity tags (the same general structure as the input data for Replace User Data), for each requested data type.
+If no chunks for a requested data type exist, an implementation MAY omit that data type from the response.
+
+To transform the public data from the output to Avro binary records, a consumer should apply the following algorithm to each data type included:
+* Determine the relevant encryption algorithm, compression codec, and object schema from the user data type and version noted.
+* For each chunk,
+  * If encryption is indicated, decrypt the chunk data using the user's secret key as in the [libsodium](https://doc.libsodium.org/public-key_cryptography/sealed_boxes) function `crypto_box_seal_open`.
+  * If compression is required, uncompress the chunk data using the specified codec.
+  * Append the chunk to the list of records for the data type.
```
wilwade, line 83: Can this happen before deserialization? Actually the deserialization is not specified here, so not sure which way you intended it to go.

pages/DSNP/UserData.md
```diff
@@ -0,0 +1,84 @@
...
+## Entity Tags
+
+To allow for application-layer caching and prevent race conditions between different applications attempting to update the same data, an implementation MUST provide an `etag` (entity tag) value with each chunk fetched, and update this for each replaced chunk.
+An implementation MUST fail the Replace User Data Operation if the `etag` values supplied do not match the current `etag` values for all chunks of the specified data type.
+
+## Replace User Data Operation
+
+The Replace User Data Operation takes the following parameters:
+
+* The DSNP User ID of the user who controls the data
```
wilwade, line 38: Trying out a different wording. The current one might be best, but it reads confusing to me.
```suggestion
* The DSNP User ID for the data
```

pages/DSNP/UserData.md
```diff
@@ -0,0 +1,84 @@
...
+  * Implementations MUST ensure that the principal invoking this Operation is this user or a transparent chain of delegation from the user to the principal exists.
```
wilwade, line 39: Do we need this requirement? Or more specifically do we need it here? I wonder if we can just rely that the requirement is listed as a part of the delegation section of the spec.

pages/DSNP/UserData.md
```diff
@@ -0,0 +1,84 @@
...
+* For each chunk generated, the application should then:
+  * If the data type requires compression, apply the compression codec noted.
+  * If the data type requires encryption,
+    * Retrieve the user's active (most recently announced) `keyAgreement` public key, U<sub>public</sub>.
+    If no key exists, one should be created and published as an Announcement before invoking the Operation.
+    * Create a sealed box, as in the [libsodium](https://doc.libsodium.org/public-key_cryptography/sealed_boxes) function `crypto_box_seal`, using U<sub>public</sub>.
+  * Include the previous `etag` value for the chunk. If the chunk is new, `etag` should be set to `null`.
+  If any chunks are to be deleted, they should be included in the input identified with the existing `etag` and a `null` value for the data.
+
+If the Operation is invoked successfully it MUST (synchronously) return a new set of `etag` values for each data type replaced, corresponding to the updated state of the data (with new chunks added and deleted chunks removed).
```
wilwade, line 59: `it` I'm not sure what the `it` here is. `it` will not be the implementation as Frequency or Ethereum are unable to respond with that data. It must be the DSNP library for that implementation?

That feels like a bit of an odd thing to require.

We could require that the async confirmation include that information, but I think that is all.

pages/DSNP/UserData.md
```diff
@@ -0,0 +1,84 @@
...
+Applications should not interpret this response as an indication that the operation was completed and a state change record emitted, as this typically occurs asynchronously.
+However, this strategy allows applications to make the optimistic assumption that, in due course, the DSNP system will reflect the intended changes, without needing to wait for asynchronous confirmation.
+
+If, on the other hand, an invocation of Replace User Data is rejected due to entity tag discrepancies, this indicates that the relevant data has changed since the last preimage was acquired, and the application should fetch the most current version of the relevant data with the Get User Data Operation, reapply any intended changes, and retry the operation.
```
wilwade, line 63: Only use of the word `preimage`. Not sure if it needs additional explanation or use elsewhere?

pages/DSNP/UserData.md
```diff
@@ -0,0 +1,84 @@
...
+
+The Replace User Data Operation MUST generate a [User Data Replaced Record](Records.md#user-data-replaced) containing the DSNP User Id and the set of updated user data types (but not the data itself).
+If the implementation detects that no change has occurred, it SHOULD omit the relevant unchanged data types from the state change record.
+
+# Get User Data Operation
+
+The Get User Data Operation takes the following parameters:
+* The DSNP User Id of the user who controls the data
+  * Note: While _writing_ user data is reserved for the user and any delegates, any authenticated principal can read any user's data (though it may be encrypted).
```
shannonwells, line 72: Why does a principal invoking a read-only operation need to be authenticated, particularly if there is encrypted data that the principal may not be able to decrypt?

pages/DSNP/UserData.md
```diff
@@ -0,0 +1,84 @@
...
+* The user data types (by system name) that should be retrieved.
+
+The operation returns a mapping of user data type to data chunks and entity tags (the same general structure as the input data for Replace User Data), for each requested data type.
```
shannonwells, line 75: could link to same specification here of the "Replace User Data" as I mentioned earlier.

pages/DSNP/UserData.md
```diff
@@ -0,0 +1,84 @@
...
+If no chunks for a requested data type exist, an implementation MAY omit that data type from the response.
+
+To transform the public data from the output to Avro binary records, a consumer should apply the following algorithm to each data type included:
```
shannonwells, line 78: Again I think algorithm steps need to be listed numerically and very much benefit from a flow chart.

pages/DSNP/Graph.md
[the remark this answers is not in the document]
wesbiggs: Agreed, changing to seconds.

Re: Parquet vs. Avro, can you clarify what you mean? I think Parquet is great for batches because you get native Bloom Filter support, self-contained schema metadata, and a columnar format that makes it easier to search. Those exact reasons make it a terrible choice for on-chain (or near-chain) data. As much as I'd rather support a single standard, the use cases suggest otherwise. If you mean we should optimize Parquet storage of INT64 types, it looks like that can be done with [DELTA_BINARY_PACKED = 5](https://parquet.apache.org/docs/file-format/data-pages/encodings/#a-namedeltaencadelta-encoding-delta_binary_packed--5). But if I'm misunderstanding the question let me know.

pages/DSNP/UserData.md
wesbiggs: Decision for initial schema is to stick with an uncompressed list of PRIDs; if we find a suitable alternative approach in the future, we can bump to a new schema.

pages/DSNP/Graph.md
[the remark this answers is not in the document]
wilwade: I'm just thinking of adding the additional data type to the table data for each announcement. Aka instead of assuming that everything is parquet as it sort of does now, we are clearer in each Announcement spec how it should be represented in Avro data as well.

image

pages/DSNP/Graph.md
wesbiggs: This is a compromise driven by the lack of a native unsigned type in Avro. The alternative would be to store as a fixed 8-byte value. This is better aligned with actual User Ids, but doesn't allow the Avro serialization to do any integer packing, which pushes responsibility to the compression codec, which has less context to work with.

Comparison of 300 GraphEdge entries using evenly distributed User Ids from 0 to 300,000,000:
| User Id data type | Avro binary serialization size (bytes) | Size after applying Brotli compression (bytes) |
| --- | --- | --- |
| `long` | 2,247 | 2,167 |
| `fixed` | 3,900 | 2,841 |

`fixed` is consistently about 30% larger after compression.

pages/DSNP/UserData.md
wesbiggs: I will clarify ("Append the chunk" is not accurate anymore). The data after the uncompress step is a sequence of Avro binary objects, so can be deserialized into logical records after that.

pages/DSNP/UserData.md
wesbiggs: Going to change this to `A DSNP User Id` because I think it should be obvious from the context (and the following line about delegation).

pages/DSNP/UserData.md
wesbiggs: Hmm, this brings up the fact that at the DSNP level we're only specifying delegation for Announcements. Let's discuss.

pages/DSNP/UserData.md
wesbiggs: I think of "the implementation" as inclusive of both the blockchain and the "adapter layer" (DSNP on Frequency). For example, Frequency will be defining in its compliance documentation the RPCs used to invoke DSNP Operations. Such an RPC could handle returning these values synchronously (which I think is the plan from discussion with @aramikm and @JoeCap08055). I can definitely change "it" to "the implementation" for clarity.

pages/DSNP/UserData.md
wesbiggs: Good callout, will change.

pages/DSNP/UserData.md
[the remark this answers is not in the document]
wesbiggs: Adding a comma for now...

pages/DSNP/UserData.md
```diff
@@ -0,0 +1,84 @@
...
+`curve25519xsalsa20poly1305` (that is, X25519 key exchange, XSalsa20 encryption, and Poly1305 message authentication) is the default authenticated encryption algorithm used in the [NaCl](https://nacl.cr.yp.to) ("Salt") library, and its successor [libsodium](https://libsodium.org). In the specification of cryptographic operations below, relevant methods from these libraries are noted. While these specific libraries are not required for DSNP compatibility, they are highly recommended.
+
+## Data Chunks
+
+Because blockchain systems often have specific limits to the amount of data that can be included in a given transaction, operations on user data deal with the data in discrete chunks.
+As implementation strategies may vary, implementations MUST define their own maximum chunk size in bytes to be used in the operations described below.
+
+## Entity Tags
+
+To allow for application-layer caching and prevent race conditions between different applications attempting to update the same data, an implementation MUST provide an `etag` (entity tag) value with each chunk fetched, and update this for each replaced chunk.
```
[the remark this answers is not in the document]
wesbiggs: This _is_ the specification, so I'll take that as suggesting that we need a more detailed description. (But yes, it is modeled after HTTP.)

pages/DSNP/UserData.md
wesbiggs: Agreed, will remove.

pages/DSNP/UserData.md
wesbiggs: Thanks, I put it in numeric outline format (but markdown is not as full-featured as one would want in this regard).

pages/DSNP/Graph.md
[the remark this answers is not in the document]
wesbiggs: OK I see what you mean. Let's break that out to a separate issue for discussion.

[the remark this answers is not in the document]
wesbiggs: See https://github.com/LibertyDSNP/spec/issues/235

pages/DSNP/UserData.md
wesbiggs: Just a bump to highlight the suggestion for `brotli` as the codec here. It looks like there was some previous discussion of using/allowing brotli with parquetjs here: https://github.com/LibertyDSNP/parquetjs/pull/17.

Some experimentation found that the brotli codec consistently gave the best compression, though not all that much more than DEFLATE, and neither gives you too much on top of Avro's native ZigZag encoding. DEFLATE is older and widely preinstalled (e.g. multiple js libraries, built into the java.util.zip package, etc.). I don't have a strong preference.

Sample output for a list of 300 GraphEdges, assuming User Ids randomly distributed between 0 and 300M:
| Format | Size (bytes) | Total Compression |
| --- | --- | --- |
| In-memory (2 x 8-byte longs) | 4800 |
| Avro binary | 2275 | 52.6% |
| Avro+DEFLATE | 2227 | 53.6% |
| Avro+brotli | 2167 | 54.9% |

pages/DSNP/Graph.md
```diff
@@ -2,29 +2,124 @@
...
+Applications are encouraged to verify connection relationships by checking to see that they are reciprocated, with the appropriate identifier present in both users' data.
+
+PRId objects MUST comply with the following [Avro](https://avro.apache.org) schema:
+
+```
+{
+    "namespace": "org.dsnp.userdata",
+    "name": "PRId",
+    "type": "fixed",
+    "size": 16,
```
wilwade:
```suggestion
    "size": 8,
```

pages/DSNP/UserData.md
```diff
@@ -0,0 +1,84 @@
...
+An implementation MUST fail the Replace User Data Operation if the `etag` values supplied do not match the current `etag` values for all chunks of the specified data type.
+
+## Replace User Data Operation
+
+The Replace User Data Operation takes the following parameters:
+
+* The DSNP User ID of the user who controls the data
+  * Implementations MUST ensure that the principal invoking this Operation is this user or a transparent chain of delegation from the user to the principal exists.
+* A map containing the set of data types to update as the keys, and tuples consisting of (1) the schema version used to encode the data type, and (2) a list where each element includes a data chunk and its associated entity tag, as the values.
```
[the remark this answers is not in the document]
wesbiggs: I had a version where I spec'ed it all out with JSON but I felt that to do so was overspecifying, since we don't elsewhere talk about how to express parameters for an operation. That said, if we want to consider it more like a "document" that is being submitted, maybe that puts a twist on it. I'm on the fence.

We could include a JSON format and just make it clear that it's non-normative.

pages/DSNP/UserData.md
```diff
@@ -0,0 +1,92 @@
...
+
+To allow for application-layer caching and prevent race conditions between different applications attempting to update the same data, an implementation MUST provide a non-empty `etag` (entity tag) value with each chunk fetched, and update this for each chunk replaced.
+
+An `etag` is any ASCII-encoded string with a minimum length of one byte and a maximum length of 255 bytes.
+
+An implementation MUST fail the Replace User Data Operation if the `etag` values supplied do not match the current `etag` values for all chunks of the specified data type.
+
+## Replace User Data Operation
+
+The Replace User Data Operation takes the following parameters:
```
wesbiggs, line 44: Need to add a key reference/index to help establish the current `keyAgreement` key used to aid in later decryption (here and in Get User Data).

pages/DSNP/UserData.md
wesbiggs: I'll change the Define Delegation Operation to say it defines both the operations and the announcement types that are delegated.

pages/DSNP/Graph.md
```diff
@@ -2,29 +2,126 @@
...
+            "name": "since",
+            "type": "long",
+            "doc": "Timestamp in Unix epoch seconds when this relationship was originally established"
+        }
+    ]
+}
+```
 
-A "follow" is the act of publicly following a [DSNP User Identity](Identifiers.md#dsnp-user-id), which results in adding this DSNP User Id to a user's social graph.
+## User Data Sets
```
wilwade, line 46: Instead of mentioning data type information here, should that just move over to `UserData.md` and reference a "See [User Data] for details"?

pages/DSNP/Graph.md
```diff
@@ -2,29 +2,126 @@
...
-### Friendship
+### Public Follows
 
-There is currently no concept of "friendship" within the DSNP network.
-Friendship requires a mutual acknowledgment between two different DSNP User Identities.
-Friendship could be thought of as "mutual following" - where two DSNP User Identities are following each other.
+Each public follow records a unidirectional relationship where one user (the controller of the graph) is interested in following the public speech of another user.
+When follows are public, they are visible to any other user or system, including the user being followed.
+
+The `publicFollows` user data type is composed of a set of GraphEdge objects, which is then compressed.
```
wilwade, line 54: Lines like this one feel out of place without additional details here, but those details are over on UserData

pages/DSNP/Graph.md
```diff
@@ -2,29 +2,126 @@
...
+
+### Private Follows
+
+A private follow also records an interest in following the public speech of another user, but the data is stored in encrypted form, meaning the relationship is not visible to the followed user, or any other user, without access to the decryption key.
```
wilwade, line 58: Simplification of English suggestion:
```suggestion
A private follow records an interest in following the public speech of another user, but the data is stored in encrypted form.
The relationship is not visible to the followed user or any other user, without access to the decryption key.
```

pages/DSNP/Graph.md
```diff
@@ -2,29 +2,126 @@
...
+
+The `privateFollows` user data type is composed of a set of GraphEdge objects, which is then compressed and encrypted.
+
+### Private Connections
+
+Private connections represent one half of a bidirectional relationship between two DSNP Users.
```
wilwade, line 64: Do we need to talk at all about why we spit the representation?

pages/DSNP/Graph.md
```diff
@@ -2,29 +2,126 @@
...
+An encrypted data set controlled by a DSNP user records which users they recognize as connections.
+A second, public, data set enables users to signal their connection to one another via Pseudonymous Relationship Identifiers.
```
wilwade, line 66: The mention here of "second" feels like it references the statement about "half" above it.

Perhaps a rewording like this?:

```suggestion
Private connections are stored in on both sides of the bidirectional relationship between two DSNP Users.
Each user can then update the status of the relationship without loss of privacy.
Each DSNP user maintains and controls two sets of data:
1. An encrypted data set that records which users are recognized as connections.
2. A public data set that is to signal connection status via Pseudonymous Relationship Identifiers.
```

pages/DSNP/UserData.md
```diff
@@ -0,0 +1,211 @@
...
+    ]
+  }
+```
+
+## Get User Data Operation
+
+The Get User Data Operation takes the following parameters:
+* The DSNP User Id of the user who controls the data
+  * Note: While _writing_ user data is reserved for the user and any delegates, anyone on the network can read any user's data (though it may be encrypted).
+* The user data types (by system name) that should be retrieved.
```
wilwade, line 151: I think we need to use User Data Types as a term.

wilwade (commented):
Got through 95% of it. Looking good. Getting down to just English issues now.

pages/DSNP/UserData.md
```diff
@@ -0,0 +1,211 @@
...
+    1. If the data type requires compression, apply the compression codec noted.
+    1. If the data type requires encryption,
+        1. Retrieve the user's active (most recently announced) `keyAgreement` public key, U<sub>public</sub>.
+  The `keyId` in the announcement should match the key identifier provided for this Operation.
+  If no key exists, one should be created and published as an Announcement before invoking the Operation.
+        1. Create a sealed box, as in the [libsodium](https://doc.libsodium.org/public-key_cryptography/sealed_boxes) function `crypto_box_seal`, using U<sub>public</sub>.
+        1. Include the previous `etag` value for the chunk. If the chunk is new, `etag` should be set to `null`.
+  If any chunks are to be deleted, they should be included in the input identified with the existing `etag` and a `null` value for the data.
+
+If the Operation is invoked successfully the implementation MUST (synchronously) return a new set of `etag` values for each data type replaced, corresponding to the updated state of the data (with new chunks added and deleted chunks removed).
```
shannonwells, line 70: If the Operation is successful is a synchronous return actually required? If so, "synchronously" being in parens is a little odd.

pages/DSNP/Graph.md
```diff
@@ -2,29 +2,126 @@
...
+{
+    "namespace": "org.dsnp.userdata",
+    "name": "GraphEdge",
+    "type": "record",
+    "doc": "A relationship to another DSNP user",
+    "fields": [
+        {
+            "name": "userId",
+            "type": "long",
+            "doc": "DSNP User Id of object of relationship"
```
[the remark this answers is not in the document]
shannonwells: How about "the other user's DSNP Id", since that is sort of what the "doc" field says?

pages/DSNP/Graph.md
shannonwells: could always link to "GraphEdge"

pages/DSNP/Graph.md
```diff
@@ -2,29 +2,126 @@
...
+* Timestamp when relationship was created, in seconds since the [Unix epoch](https://en.wikipedia.org/wiki/Unix_time)
 
-Implementation coming in the future. See the [DSNP Whitepaper](https://github.com/LibertyDSNP/papers) for foundational ideas.
+GraphEdge objects MUST conform to the following [Avro](https://avro.apache.org) schema:
 
-### Following
+```
+{
+    "namespace": "org.dsnp.userdata",
+    "name": "GraphEdge",
```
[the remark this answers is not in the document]
wilwade: Let's try a separate page just for Avro schemas and see if it works.

pages/DSNP/Graph.md
wilwade: Link to the "User Data Types" table

pages/DSNP/Graph.md
```diff
@@ -2,29 +2,126 @@
...
+
+The following section describes how these values are used in the PRId generation process.
+
+#### Algorithm
+
+In the following section, the Alice to Bob identifier for context C is called PRId<sub>A→B,C</sub>, and the corresponding Bob to Alice identifier is called PRId<sub>B→A,C</sub>.
+
+A PRId is derived from Alice and Bob's `keyAgreement` key pairs, using a key exchange protocol as follows. To illustrate the cryptographic operations required, the relevant functions from [libsodium](https://libsodium.org) are noted. Sodium is a stable, fast, free, and cross-platform cryptography library, and supports all encryption algorithms used in the DSNP specification out of the box.
+
+1. Both Alice and Bob generate an asymmetric key pair, and each publishes a Public Key Announcement with their generated X25519 public key and a `keyType` value of `keyAgreement`.
```
wilwade, line 111: @shannonwells to try to sketch something out for this

pages/DSNP/UserData.md
wilwade: Answer: Stick with DEFLATE

pages/DSNP/UserData.md
```diff
@@ -0,0 +1,211 @@
...
+* A [Key Identifier](Types/PublicKey.md#keyid) for the `keyAgreement` key pair used to encrypt any private data in the operation.
+  (If only unencrypted user data types are included, the key identifier is optional.)
+* A map containing the set of data types to update as the keys, and tuples consisting of (1) the schema version used to encode the data type, and (2) a list where each element includes a data chunk and its associated entity tag, as the values.
+
+If the Operation is successful, any previous data associated with the user for each data type included in the input MUST be removed and replaced by the new data.
+
+Data chunks should be generated for each included data type using the following sequence of operations:
+1. Serialize the data records, in Avro binary format, according to the versioned schema.
+1. Divide the data records into one or more chunks, with each chunk containing a maximum byte count specified by the implementation.
+* Note that the combined chunks constitute a different, and smaller, binary than a complete Avro file, as they exclude the header information, schema, and any block-level metadata.
```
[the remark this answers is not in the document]
wilwade: Add to the Glossary?

pages/DSNP/UserData.md
wilwade:
```suggestion
If the Operation is invoked successfully the implementation MUST synchronously return a new set of `etag` values for each data type replaced, corresponding to the updated state of the data (with new chunks added and deleted chunks removed).
```

pages/DSNP/UserData.md
```diff
@@ -0,0 +1,211 @@
...
+  Each chunk should start and end with a complete Avro object; in other words, chunk boundaries MUST NOT occur mid-record.
+    Applications SHOULD try to include as many records as possible within the byte length limit, but because compression and encryption may alter the byte size versus the raw Avro binary records, they may wish to use a heuristic approach that sometimes produces non-optimal chunks.
+3. For each chunk generated, the application should then:
+    1. If the data type requires compression, apply the compression codec noted.
+    1. If the data type requires encryption,
+        1. Retrieve the user's active (most recently announced) `keyAgreement` public key, U<sub>public</sub>.
+  The `keyId` in the announcement should match the key identifier provided for this Operation.
+  If no key exists, one should be created and published as an Announcement before invoking the Operation.
+        1. Create a sealed box, as in the [libsodium](https://doc.libsodium.org/public-key_cryptography/sealed_boxes) function `crypto_box_seal`, using U<sub>public</sub>.
```
aramikm, line 66: so seal box would need creation of an ephemeral key, do we need to include that or it is already assumed since we are using `crypto_box_seal`? 

On another topic do we need another layer of key derivation with different context for encryption or is that an overkill?

pages/DSNP/Graph.md
```diff
@@ -2,29 +2,126 @@
...
+Bob can also derive the PRId subkey for Alice's DSNP User Id and encrypt Alice's User Id, using his own as the nonce, to generate the Bob-to-Alice PRId, and then publish it to his own list, if desired.
+
+If Alice or Bob wants to prove to a third party that their PRIds are in each other's `privateConnectionPRIds` list, they can provide the third party with their own subkey <code>k<sub>Alice</sub></code> or <code>k<sub>Bob</sub></code>. The third party can repeat the encryption step using Alice and Bob's User Ids, and check that the output is present in the published set of PRIds. The shared secret `k` (used as a master key in this algorithm) should _not_ be divulged.
+
+In more formal terms:
+* <code>PRId<sub>A→B,C</sub></code> = <code>XSalsa20(HKDF(SharedSecret<sub>AB</sub>, _DSNP User Id of B_, "PRIdCtx<i>C</i>"), _DSNP User Id of B_, _DSNP User Id of A_)</code>
+* <code>PRId<sub>B→A,C</sub></code> = <code>XSalsa20(HKDF(SharedSecret<sub>AB</sub>),_DSNP User Id of A_, "PRIdCtx<i>C</i>"), _DSNP User Id of A_, _DSNP User Id of B_)</code>
```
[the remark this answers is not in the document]
wilwade: @aramikm See if this makes it more clear: https://github.com/LibertyDSNP/spec/pull/238

pages/DSNP/UserData.md
```diff
@@ -0,0 +1,211 @@
...
+		"keyId": integer
+      },
+      {
+        "data": base64_string,
+        "etag": string,
+		"keyId": integer
+      }
+    ]
+  },
+  "privateConnectionPRIds": {
```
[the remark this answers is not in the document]
wilwade: Good point. That is the old version in this example. It needs to be updated.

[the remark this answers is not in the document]
wesbiggs: That's not possible without a lot of rework here.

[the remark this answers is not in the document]
wesbiggs: It was going to be up to the Frequency SDK/adapter to make sure it got these at the same time, but they are separate user data types from a DSNP perspective.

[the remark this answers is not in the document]
wilwade: That breaks the etag however for the chunk for privateConnections because that chunk includes this data.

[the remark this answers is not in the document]
wesbiggs: Generation and maintenance of the etag is up to the implementation (DSNP does not know that it's a hash) so I think this is OK.

pages/DSNP/UserData.md
```diff
@@ -0,0 +1,211 @@
...
+Applications should not interpret this response as an indication that the operation was completed and a state change record emitted, as this typically occurs asynchronously.
+However, this strategy allows applications to make the optimistic assumption that, in due course, the DSNP system will reflect the intended changes, without needing to wait for asynchronous confirmation.
+
+If, on the other hand, an invocation of Replace User Data is rejected due to entity tag discrepancies, this indicates that the relevant data on the network has changed since the entity tags were acquired by the application, and the application should fetch the most current version with the [Get User Data Operation](#get-user-data-operation), reapply any intended changes, and retry the operation.
+
+The Replace User Data Operation MUST generate a [User Data Replaced Record](Records.md#user-data-replaced) containing the DSNP User Id and the set of updated user data types (but not the data itself).
+If the implementation detects that no change has occurred, it SHOULD omit the relevant unchanged data types from the state change record.
```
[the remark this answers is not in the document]
wesbiggs: Line 75 mentions retrying. I think that's enough for the DSNP side, the Frequency extrinsics can go into more detail.

pages/DSNP/UserData.md
wesbiggs: I added more description for "sealed box", hopefully that helps.

For the second comment, if I'm understanding it correctly, yes, we will add an additional context for PRIds used for message encryption (just not part of this PR).

wilwade (approved):
🎉

shannonwells (approved):

PR #248, opened 2023-04-07
Remove revokedAsOf from Public Key Announcement

wesbiggs:
Problem
=======
The semantics of the `revokedAsOf` field in the pre-release specification for Public Key Announcement are unclear, and may at this point cause more confusion than the problems it is intended to solve.

Solution
========
This PR  removes the `revokedAsOf` field from the Announcement. Future evolution of the specification may propose a different way of capturing user intent around declaring that a key has been compromised or otherwise should be considered untrustworthy, either entirely or from a specific date and time.

From discussion with @wilwade @aramikm @JoeCap08055 

Change summary:
---------------
- [ ] Removed field and relevant parts of description
- [ ] Clarified format for publicKey

Steps to Verify:
----------------
1. Generate site and review

### DIFF
--- a/.yaspellerrc.json
+++ b/.yaspellerrc.json
@@ -75,6 +75,7 @@
     "deadDropId",
     "dequeued",
     "deserialized",
+    "deterministically",
     "dsnpUserId",
     "ecrecover",
     "emoji[s]?",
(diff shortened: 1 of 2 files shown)


JoeCap08055 (approved):

aramikm (approved):
Looks good to me!

PR #252, opened 2023-04-26
Fix for missed hash description

wesbiggs:
Clean up a straggling reference to keccak hashing and simplify announcement key descriptions.

Problem
=======
There was an un-updated reference to pre-multihash hashing on the Identifiers page.

Solution
========
Moved the details of hash definition and supported algorithms to the Identifiers page and reworked links.


wilwade (approved):

shannonwells (approved):
nice

PR #254, opened 2023-05-01
Add required icon hash fields to Profile example

wesbiggs:
Problem
=======
Image Link objects in DSNP require a hash, but the example given for icons in an Activity Content Profile did not give these.

Solution
========
```
go install github.com/multiformats/go-multihash/multihash@latest
curl -s https://placekitten.com/64/64 | multihash -a sha2-256
curl -s https://placekitten.com/256/256 | multihash -a sha2-256
```

Somewhat ironically, however, the kittens returned by placekitten.com are nondeterministic, so YRMV. If we feel the example hashes should be fully reproducible, we should use more deterministic kittens, or perhaps try puppies instead.

Change summary:
---------------
- [ ] Updated Activity Content Profile example


wilwade (approved):

shannonwells (approved):
I think it's fine.

Issue #253, opened 2023-05-01
DIP-253 Specify usage of "actor" in DSNP-linked Activity Content


wesbiggs:
# Abstract

By specifying how to use the "actor" key in Activity Content (Note and Profile) referenced from DSNP Announcements, we can achieve better trust in the authenticity of the provenance of the content.

# Motivation

The current DSNP Activity Content specification does not mandate the use of "actor" to indicate the originator of a piece of content (in DSNP, a Note or Profile). Because of this, it would be possible for Bob to create an Announcement referencing the same Activity Content item that Alice has published, thus claiming it as his own. Alice may already have posted the Announcement, or Bob might be able to execute a form of man-in-the-middle attack where he front-runs the Announcement, claiming afterward that _Alice_ was the plagiarist. While this scenario may be unlikely for the vast majority of user-generated content, preserving provenance is important when dealing with sensitive, controversial, or novel content.

# Specification Pull Request

The specification will be updated to encourage the use of the DSNP User URI in Note and Profile Activity Content objects as the actor key. The key (and its usage) can remain optional; there may still be reasons why users want to reference content created using non-DSNP systems, or content that is meant to be Announced by multiple users.

For example,
`"actor": "dsnp://1311768467294899700"`

The spec should mandate that applications treat as invalid (or at least warn when encountering) any content where the actor is a DSNP User URI but does not match the `senderId` indicated in the corresponding Announcement.

PR: #255 

# Rationale

The "actor" key is already defined in the W3C Activity Content spec and it seems intended to have semantics that can be contextually defined.

An alternative approach would be to add a digital signature to the Activity Content. This has several disadvantages: (1) It requires a user (or their user agent/application) to have access to their signing key with every broadcast or reply; this then requires a tradeoff between security and usability. (2) It's unnecessary; there is already a signature or delegation signature chain leading to the Note/Profile. (3) There is no single common approach for adding signatures to JSON data; various approaches either end up encoding the entirety of the Note to form an envelope, which wastes space, or requiring a priori information about how to remove the signature in order to compute the original content hash. Other approaches require the signature to be embedded in HTTP headers, which (while still compatible with this approach) would place an additional technical burden on web hosts providing access, and may not be compatible with future content access protocols (e.g. IPFS) or HTTP gateways to the same.

[...]

PR #255, opened 2023-05-02
DIP-253 Add actor specification to Activity Content [DRAFT FOR DISCUSSION]

wesbiggs:
Problem
=======
In the current spec it is possible for a user to announce (Broadcast, Reply, Profile) an Activity Content document that another user created and thereby claim it as their own. We have content authentication (contents of URL must match the hash) but not content provenance (user who claims to have originated the content must match the sender of the Announcement).

Link to GitHub Issue(s): https://github.com/LibertyDSNP/spec/issues/253 has further background.

Solution
========
We include the `actor` key (loosely defined in the Activity Content 2.0 specification) in the DSNP Activity Content specification, and define it as representing a DSNP User if it contains a DSNP User URI. By denormalizing in this way we enable readers to verify the linkage between `senderId` and `actor`.

Change summary:
---------------
- [ ] Updated Spec Versions (pre-1.2.1) for Activity Content and DSNP specs.
- [ ] Added `actor` to Note and Profile Activity Content objects
- [ ] Added info on the Announcements page describing the trust semantics of Activity Content

Steps to Verify:
----------------
1. Generate and view spec site.

### DIFF
--- a/.spellcheckerdict.txt
+++ b/.spellcheckerdict.txt
@@ -65,6 +65,7 @@ PNG
 Polkadot
 Poly1305
 pre-configured
+Prerelease
 PRId([ABs])?
 publicKey
 repo
--- a/pages/ActivityContent/Overview.md
+++ b/pages/ActivityContent/Overview.md
@@ -1,5 +1,5 @@
 # Activity Content Specification
-__Version 1.2.0__
+__Version pre-1.2.1__
 
 Content references shared via the DSNP consist of URLs pointing to documents containing Activity Streams JSON objects.
 For the purposes of the DSNP, restrictions are placed on the [Activity Streams 2.0](https://www.w3.org/TR/activitystreams-core/) specification.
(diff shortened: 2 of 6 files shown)


wesbiggs:
> I have a vague recollection that this was done intentionally to allow reposts but I'm not certain. I do wonder how reposts would work now.

I wasn't here for those prior discussions but my opinion is that reposting should be explicit, not implied, and that we'd want a new announcement type that allowed this with a reference to the original announcement by its DSNP Content URL (like inReplyTo for Reply/Reaction) rather than linking to the external Activity Content directly.

The other option is that instead of saying that announcing with a URL pointing to someone else's Note is invalid, we say that it should be treated as if it was a repost in that case. This works if there is an `actor` field on the Note (i.e. you can know whose content is being reposted) but would require readers/applications to keep a reverse lookup table from the Note URL to its Announcement in order to authenticate the original actor.

wilwade:
Notes from the DSNP Community Call 2023-06-22
- Does not add greater authorship guarantees
- Does not impact privacy either way
- Signatures would provide a better solution when high degrees of authorship is needed.

PR #258, opened 2023-08-22
fixed typo  activity content

aramikm:
Problem
=======

1. `Update` announcement type should be included in `Note` Activity content table
2. in `Note` activity content example there was a `summary` field which is not part of acceptable fields

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 .vscode
 node_modules/
 book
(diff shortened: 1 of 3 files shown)


wesbiggs (approved):
Great, thanks @aramikm!

wilwade (approved):
💯

PR #269, opened 2023-12-18
Fix a nonstandard example DSNP URI

wesbiggs:
Old: `dsnp://0x12345678`
New: `dsnp://12345678`

I checked that this was not being repeated anywhere else in the spec.


wilwade (approved):
Good find

Issue #260, opened 2023-10-30
Glossary needs to be updated


shannonwells:
### Steps to reproduce
The [DSNP Spec Glossary](https://spec.dsnp.org/Reference/Glossary.html) has a bunch of outdated and overly specific references, such as 'Announcement - An item of data, typically an image or note, posted to the blockchain via a "Batch Publication"' and 'Identity Contract - The smart contract used to store a user's delegated public addresses and permissions, see "Contract address"', which should be updated to reflect current reality.

We may wish to add more definitions.

### Expected behavior
* We no longer have an Identity Contract; that was the POC implementation on Ethereum.
* Announcements don't have to be posted in batches nor do they have to be posted to a blockchain.

PR #270, opened 2024-01-04
DIP-260 Remove outdated entries from the glossary, and clarify others

wesbiggs:
Problem
=======
Glossary contains out-of-date terminology no longer associated with the spec.

Link to GitHub Issue(s): #260 

Solution
========
Removed items that clearly no longer applied.
Edited a few others.

The brevity of the page after these changes arouses suspicion that we may not benefit much from having a glossary, but perhaps we will find items to add at a later time.

Change summary:
---------------
- Changes to Glossary page only. I do not believe this warrants a spec versioning bump.

### DIFF
--- a/pages/Reference/Glossary.md
+++ b/pages/Reference/Glossary.md
@@ -1,16 +1,10 @@
 # Glossary
 
-* **Activity Content** - The linked content connected to an "Announcement"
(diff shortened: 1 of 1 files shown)


wilwade (approved):

shannonwells (approved):
:shipit:

PR #271, opened 2024-02-29
Update GitHub actions to all be Node20+

wilwade:
What it says on the can.

Has the same code as https://github.com/LibertyDSNP/frequency-docs


wesbiggs (approved):
LGTM

Issue #263, opened 2023-11-20
DIP-263 Use User Data for public keys


wesbiggs:
# Abstract

Public keys should be treated as User Data types rather than content announcements.

We remove the Public Key Announcement and replace with a User Data type for each key type, namely `keyAgreement` and `assertionMethod`. These formats will be defined in Avro schema (a single Avro type, PublicKey), and added to the User Data section of the spec. The Public Key Announcement will be moved to "Migrated Announcements".

# Motivation

Public keys (for `keyAgreement` and `assertionMethod` key types, at least) are important for DSNP participants to be able to retrieve deterministically and quickly in order to accomplish tasks like graph encryption/decryption and credential verification.

If these are implemented as announcements (as is currently implied by the specification), the consumer must either rely on an outside content index, or an archive of state change events. The former is subject to man-in-the-middle attacks if not maintained directly by the querying party, and retrieval of the latter may be highly inefficient.

By defining these as User Data items, we guarantee that they can be accessed directly by DSNP Operations.

# Specification Pull Request

Current change pull request: #276 

# Rationale

This proposal is informed by implementation choices made by the DSNP over Frequency implementation.

# Backwards Compatibility

This is a declarative change only, so there should be no impact on current DSNP systems provided they are using a compatible storage strategy today.

# Reference Implementation and/or Tests

DSNP over Frequency is proposed as a reference implementation.

# Security Considerations

Treating keys as User Data should help avoid security challenges that might become a problem if content indexers were used:

1. Content indexers cannot be fully trusted to maintain data integrity as they sit outside the consensus system.
2. Content indexers could be subject to man-in-the-middle attacks.
3. Content indexers that are not fully in sync with the latest updates to keys could give the wrong answer to queries for a user's most recent key.

# Dependencies

None.

# References

DSNP spec:
- [Public Key Announcement](https://spec.dsnp.org/DSNP/Types/PublicKey.html)
- [User Data](https://spec.dsnp.org/DSNP/UserData.html)

# Copyright

Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/).

wesbiggs:
For discussion: 

Should public keys include proof of ownership verifiable by consumers?

For assertionMethod keys, proof would consist of signing the DSNP User Id with the private key.

For keyAgreement keys, is this possible? (avoiding replay attacks)

wesbiggs:
[...]

PR #276, opened 2024-04-11
DIP-263 Use User Data for Public Keys

wesbiggs:
Problem
=======
See #263 

Solution
========
- Moved old Public Key Announcement to "Migrated Announcements".
- Added Avro type PublicKey.
- Updated User Data section to define key agreement and assertion method key types.


pages/DSNP/UserData.md
```diff
@@ -190,12 +191,12 @@ The following example illustrates the output of a Get User Data Operation invoca
       {
         "data": base64_string,
         "etag": string,
-        "keyId": integer
+        "keyIndex": integer
```
wilwade, line 194: Would this be a change to the https://github.com/LibertyDSNP/graph-sdk/ that needs to happen?

pages/DSNP/Announcements.md
```diff
@@ -17,9 +17,11 @@ Each Announcement has an enumerated type for use when separating out a stream of
 | 4 | [Reaction](Types/Reaction.md) | a public visual reply to a Broadcast | no | no |
 | 5 | [Profile](Types/Profile.md) | a profile | YES | no |
 | 6 | [Update](Types/Update.md) | an update to content| YES | no |
-| 7 | [Public Key](Types/PublicKey.md) | a public key for secure communication | no | no |
+| 7 | ~~[Public Key](Types/PublicKey.md)~~<sup>b</sup> | ~~a public key for secure communication~~ | ~~no~~ | ~~no~~ |
```
wilwade, line 20: This, or just an update, should likely also happen inside the Frequency System spec

pages/DSNP/UserData.md
wesbiggs: I don't think this "example" JSON serialization is used directly in graph-sdk, but @aramikm or Joe might know better. I'll get some more eyes on it. If there's prior art out there using `keyId` I'd be fine with keeping it that way for simplicity's sake, but documenting that `keyId` is the index.

wesbiggs: Per Joe this won't impact graph-sdk, so resolving this thread.

pages/DSNP/Announcements.md
wesbiggs: Added Frequency page changes, please review.

pages/DSNP/UserData.md
wilwade: After thinking about this, I think this should just keep `keyId`. It is an identifier even if it applies an index algo.

wesbiggs: Agreed

wilwade (approved):
Just updating my approval for the small additional updates.

PR #284, opened 2024-09-18
Remove Rococo and add DSNP 1.3 mappings

wesbiggs:
Remove Rococo, update DSNP 1.3 mappings, errata for public-key-key-agreement multicodec confusion

Still some TBDs on schema Ids, tentatively trying to get those in before we merge this.


shannonwells:
Should this be a draft until the IDs are settled?

PR #291, opened 2024-10-03
Minor hygiene items

wesbiggs:
- open external links in a new tab (some via mdbook extension, others by hardcoding HTML)
- fix reference to deprecated public key announcement in PRId algorithm
- give a more useful internal link for assertionMethodPublicKeys in the Verifiable Credentials overview
- name items in the PRId description a little more consistently

### DIFF
--- a/.github/workflows/verify-pr.yaml
+++ b/.github/workflows/verify-pr.yaml
@@ -24,7 +24,9 @@ jobs:
         run: npm run lint
 
       - name: Set up mdBook 📚
-        run: cargo install mdbook mdbook-linkcheck mdbook-external-links
+        uses: ./.github/workflows/common/set-up-mdbook
+        with:
+          token: ${{secrets.GITHUB_TOKEN}}
 
       - name: Build mdBook 📚
         run: mdbook build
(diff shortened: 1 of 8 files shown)


shannonwells (approved):
Changes look good. Reviewed code only, did not download the branch.

wesbiggs:
Ah, `jontze/action-mdbook` doesn't support the `mdbook-external-links` plugin. Will need to set up the Github action in a different way if we want to use that for running checks/deployments.

.github/workflows/verify-pr.yaml
```diff
@@ -24,9 +24,7 @@ jobs:
         run: npm run lint
 
       - name: Set up mdBook 📚
-        uses: ./.github/workflows/common/set-up-mdbook
```
wilwade, line 27: If you are removing this, then you can also remove `./.github/workflows/common/set-up-mdbook`

wesbiggs: I changed the common script instead. Running cargo install is slower, but more aligned with local usage, and this repo is not so time-sensitive.

wilwade (approved):

Issue #301, opened 2025-09-09
Lock NPM Packages


shannonwells:
Pin package.lock dependencies to exact versions, i.e. remove any "~" or "^" in front of versions.

PR #302, opened 2025-09-10
Pin dependencies and fix broken links

wesbiggs:
Problem
=======
Per #301, to avoid supply chain attacks, update `package.json` to pin exact versions of packages.

Solution
========
- set the versions to the exact values from `npm list`
- fix a couple of broken links that were discovered when rerunning the build


shannonwells (approved):
I reviewed the code changes in the files, and I checked that the new links go to where they say they go.

